PR #4, opened 2026-04-27
improve gateway description in README

oraNod:
## Description

Changes the description of gateway in the README to be more meaningful outside of the downstream context.

## Type of Change
- [x] Documentation update

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Jewel
 
-Jewel provides the source code for the Gateway API used in Red Hat Ansible Automation Platform.
+Jewel provides the source code for the Gateway API used by Ansible services built as Django Ansible Base (DAB) applications.
 
 
 > The project name, Jewel, refers to a precious stone known as the Eye of the Sea from the same novel that originated the word Ansible, Ursula K. Le Guin's *Rocannon's World*.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Jewel
 
-Jewel provides the source code for the Gateway API used in Red Hat Ansible Automation Platform.
+Jewel provides the source code for the Gateway API used by Ansible services built as Django Ansible Base (DAB) applications.
```
oraNod: Not sure if this phrasing is too wordy. Maybe we can just say "used by Django Ansible Base (DAB) services."?

tznamena: @oraNod There is more to "gateway" than just APIs. 
Also, it is not the services that use gateway, the gateway is the component that routes traffic to other services. 
Lastly, I would leave out the part about DAB because it introduces other components at the first line of readme.

oraNod: Thanks @tznamena I've rephrased this in a new commit. Hopefully it reads a bit better. Trying to keep it simple yet descriptive.

bhavenst (approved):

tznamena (approved):

PR #6, opened 2026-04-28
Add support for workflow_dispatch

tznamena:
## Description

- What is being changed?
  - added additional trigger to back-port workflow
- Why is this change needed?
  - with current implementation, the workflow could not be triggered via GH CLI
- How does this change address the issue?
  - with both workflow_call and workflow_dispatch the workflow can be triggered either from another workflow or by GH CLI

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

## Testing Instructions

Testing is optional I would say, as it requires several not so trivial items to setup before testing.

### Prerequisites

1. You will need a set of branches, one that's at least one commit ahead of the other. This can be devel branch and another temporary branch that's reset to a commit before devel's latest commit. In this case, devel will be source branch, the temporary branch will be target branch. 
2. Your repository needs to be configured to allow Actions to create Pull requests (not allowed by default)
3. Your repository will need to have PGP_PRIVATE_KEY secret that contains PGP private key, base64 encoded, with no passphrase. You could re-use you existing one if you didn't set it with a passphrase.

### Steps to Test
1. Checkout the PR and push it into a branch in your fork (in the example below, it will be PR-AAP-72147)
2. Get the branch names
3. Get the commit SHA
4. Run the GH CLI like the example below:
    ```sh
    gh workflow run backport.yml -f source_branch="devel" \
    -f source_commits="f9b6f78836817947b5d2952af37de87b8201ee36" \
    -f target_branches="devel-before-readme-update" \
    -f git_committer_name="Your name goes here" \
    -f git_committer_email="youremail@somewhere.com" \
    --repo your-org-name/jewel --ref PR-AAP-72147
    ```

### Expected Results

You should get message:
> ✓ Created workflow_dispatch event for backport.yml at PR-AAP-72147

## Additional Context

You can see successful test here, the PR created by triggering the workflow with command like shown above: https://github.com/tznamena/jewel/pull/1

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,6 +1,54 @@
 name: Backport Cherry-Pick
 
 on:
+  # This workflow support two types of triggers:
+  # - workflow_dispatch: allows manual invocation via the GitHub UI or the gh CLI
+  #     e.g.  gh workflow run backport.yml -f source_commits="abc1234" -f target_branches="stable-2.6"
+  # - workflow_call: allows invocation from another GitHub Actions workflow
+  #
+  # IMPORTANT: Keep the inputs sections of both triggers in sync!
+  # They must define the same inputs with the same types and defaults.
+  # If you add, remove, or rename an input, make the same update in both triggers.
+  # Note: Secrets are only specified for workflow_call trigger.
+  workflow_dispatch:
+    inputs:
+      source_commits:
+        description: "Commit SHA(s) to backport — single SHA or comma-separated list (e.g., abc1234 or abc1234,def5678)"
+        required: true
+        type: string
+      target_branches:
+        description: "Comma-separated target branches"
+        required: true
+        type: string
+      jira_ticket:
+        description: "Jira ticket for tracking (e.g., AAP-12345)"
+        required: false
+        type: string
+        default: ""
+      source_branch:
+        description: "Branch where the commit must exist"
+        required: false
+        type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/backport.yml
```diff
@@ -1,6 +1,54 @@
...
+      source_commits:
+        description: "Commit SHA(s) to backport — single SHA or comma-separated list (e.g., abc1234 or abc1234,def5678)"
+        required: true
+        type: string
+      target_branches:
+        description: "Comma-separated target branches"
+        required: true
+        type: string
+      jira_ticket:
+        description: "Jira ticket for tracking (e.g., AAP-12345)"
```
oraNod, line 24: Can this also be set to `NO_JIRA`? I'm not sure if that would break something if there is a `JIRA_BASE_URL` variable that gets prepended to the ticket.

tznamena: This is by default an empty string so the caller does not need to pass it in and then resulting PR (to back-port the commits) will not contain link to the Jira ticket. 
Unless I'm missing something, NO_JIRA would work exactly the same as empty string does now.

oraNod: Thanks for explaining @tznamena should have noticed that empty string...

jay-steurer (approved):

huffmanca (approved):

PR #9, opened 2026-04-28
[AAP-73441] Add auto-close PR workflow for protected branches

john-westcott-iv:
## Summary
- Adds a GitHub Actions workflow that auto-closes PRs opened against configurable protected branches
- Controlled via two repository variables: `AUTO_CLOSE_PR_BRANCHES` (space-separated branch list) and `AUTO_CLOSE_PR_MESSAGE` (optional comment before closing)
- When `AUTO_CLOSE_PR_BRANCHES` is empty or unset the workflow is a no-op, so jewel is unaffected while aap-gateway can set `devel` to block direct PRs

## Test plan
- [ ] Verify workflow is skipped when `AUTO_CLOSE_PR_BRANCHES` is not set
- [ ] Set `AUTO_CLOSE_PR_BRANCHES=devel` and open a PR against devel — confirm it is closed
- [ ] Set `AUTO_CLOSE_PR_MESSAGE` and verify the comment appears before close
- [ ] Open a PR against a non-protected branch — confirm it is left open

### DIFF
--- a/.github/workflows/auto-close-pr.yml
+++ b/.github/workflows/auto-close-pr.yml
@@ -1,0 +1,45 @@
+---
+name: Auto-Close PR
+
+on:
+  pull_request_target:
+    types: [opened, reopened]
+
+permissions:
+  pull-requests: write
+
+jobs:
+  auto-close:
+    if: vars.AUTO_CLOSE_PR_BRANCHES != ''
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check if PR targets a protected branch
+        id: check
+        env:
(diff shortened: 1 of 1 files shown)


bhavenst (approved):
Does it/can it prevent/kill the other workflows?

huffmanca (approved):

john-westcott-iv:
@bhavenst I think once it closes the other workflows are killed. We could make the other workflows dependent on this one though if we wanted. So that it would run first (its pretty quick) but it would delay other workflows until its done.

PR #29, opened 2026-05-07
fix(AAP-74625): change runtime_feature_flag default to true

zkayyali812:
## Description

Closes: [AAP-74625](https://redhat.atlassian.net/browse/AAP-74625)

Changes the `RUNTIME_FEATURE_FLAGS` setting default from `False` to `True`, enabling runtime feature flag toggling for all customers by default.

This default was originally set to `False` due to concerns that runtime modifications could incorrectly override install-time flag values. That precedence issue was resolved in AAP-61836, which enforced that install-time specified values always take precedence over runtime modifications. With that fix in place, there is no longer a correctness reason to ship with runtime flags disabled.

**Upgrade impact:** Because `RUNTIME_FEATURE_FLAGS` is a `settings_bound` preference, its value is always resolved from Django settings at runtime (the database value is not consulted). Customers upgrading who did not explicitly set this setting will automatically pick up the new `True` default. Customers who explicitly set `RUNTIME_FEATURE_FLAGS=False` at install time will retain their opt-out.

### Files changed
- `aap_gateway_api/defaults.py` — Django setting default `False` → `True`
- `aap_gateway_api/registered_preferences.py` — `getattr` fallback `False` → `True`
- `aap_gateway_api/tests/conftest.py` — Test default `False` → `True`

## Type of Change
- [ ] Documentation update
- [ ] Test update
- [x] Configuration change

## Self-Review Checklist

### Prerequisites
- A running gateway development environment (`make docker-compose`)

### Steps to Test
1. Verify `GET /api/gateway/v1/settings/feature_flags/` returns `RUNTIME_FEATURE_FLAGS: true` without any explicit setting override
2. Verify `PATCH /api/gateway/v1/feature_flags/{id}/` succeeds for a run-time toggle flag (previously would return 403 unless explicitly enabled)
3. Set `RUNTIME_FEATURE_FLAGS=False` in settings and verify the opt-out still returns 403 on PATCH
4. Run the feature flag test suite: `python -m pytest aap_gateway_api/tests/feature_flags/test_feature_flag_api.py -v`

### Expected Results
- Runtime feature flag toggling works out of the box without any configuration
- Explicit opt-out via `RUNTIME_FEATURE_FLAGS=False` continues to disable runtime toggling
- All existing tests pass without modification (tests use explicit `override_settings` for both `True` and `False` cases)

### Required Actions

## Summary by CodeRabbit

* **Chores**
  * Runtime feature flags are now enabled by default across application settings and preferences, with support for settings-based overrides.
  * Test infrastructure improved with enhanced preference management utilities providing automatic value restoration and cache cleanup between test executions.


huffmanca (approved):
Discussed in standup. Changes are clean enough. Approved.

tznamena (approved):

Dostonbek1 (approved):

PR #34, opened 2026-05-08
fix(AAP-73972): fix oidc group membership

zkayyali812:
## Description

OIDC token claims `aap_organizations` and `aap_teams` incorrectly included every organization and team in the system when the authenticated user was a system administrator. This happened because `access_qs()` bypasses RBAC evaluation for superusers via `has_super_permission()`, returning the full unfiltered queryset. OIDC group claims should reflect explicit membership, not implicit superuser access — platform-level privilege is already conveyed by `aap_system_role: system_administrator`.

This change replaces `Organization.access_qs()` / `Team.access_qs()` with direct `RoleUserAssignment` queries that only return explicit role assignments. It also changes the claim format from `"role"` (singular string) to `"roles"` (sorted list) to correctly represent users who hold both admin and member assignments on the same org or team.

**Before:**
```json
{
  "aap_organizations": [
    {"name": "Default", "role": "admin"},
    {"name": "Org2", "role": "admin"},
    {"name": "admins", "role": "admin"}
  ],
  "aap_system_role": "system_administrator"
}
```

**After (superuser with no explicit assignments):**
```json
{
  "aap_organizations": [],
  "aap_teams": [],
  "aap_system_role": "system_administrator"
}
```

**After (user with both admin and member on same org):**
```json
{
  "aap_organizations": [
    {"name": "My Org", "roles": ["admin", "member"]}
  ]
}
```

Fixes: [AAP-73972](https://redhat.atlassian.net/browse/AAP-73972)

## Type of Change
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

The `role` (string) -> `roles` (list) format change is breaking for any OIDC relying party consuming these claims. The blast radius is small since the OIDC provider feature was recently added (AAP-68137).

## Self-Review Checklist

### Prerequisites
- AAP instance with multiple organizations and teams
- A system administrator user with no explicit org/team memberships
- OIDC authentication configured via AAP Gateway

### Steps to Test
1. Authenticate as a system administrator via OIDC
2. Decode the OIDC token and inspect `aap_organizations` and `aap_teams`
3. Verify they are empty arrays (not every org/team in the system)
4. Verify `aap_system_role` is `system_administrator`
5. Explicitly assign the superuser as a member of one org and one team via the API
6. Re-authenticate and verify only those explicit assignments appear in the token
7. Assign a normal user as both admin and member of the same org
8. Verify the token shows `"roles": ["admin", "member"]` for that org

### Expected Results
- Superuser with no explicit assignments: `aap_organizations: []`, `aap_teams: []`
- Superuser with explicit assignments: only those assignments appear
- User with dual roles: `roles` list contains both `"admin"` and `"member"`
- `aap_system_role` is unaffected and correctly reflects `system_administrator`, `system_auditor`, or `normal_user`

### Root cause
`oauth2_validator.py` called `Organization.access_qs(user, 'member')` and `Team.access_qs(user, 'member')` to build claims. `access_qs()` delegates to `AccessibleObjectsDescriptor` in `django-ansible-base/ansible_base/rbac/evaluations.py`, which short-circuits at line 86-87 and returns the full queryset for any user where `has_super_permission()` is true. This is correct for RBAC authorization checks but semantically wrong for OIDC group claims.

### Query optimization
Replaced 4 separate `RoleDefinition.objects.managed.*` lookups (each calling `get_or_create`) with a single `RoleDefinition.objects.filter(name__in=[...])` query. Total DB queries per `get_userinfo_claims` call: 5 (1 role definition lookup + 2 assignment queries + 2 object queries).

### Required Actions
- [x] Requires documentation updates
  
## Summary by CodeRabbit

* **Updates**
  * Enhanced user role information in authentication claims. Organizations and teams now display role assignments as lists for clearer visibility into permission levels.

* **Tests**
  * Expanded test coverage for authentication claims validation, including multi-role scenarios and system administrator/auditor behavior verification.

### DIFF
--- a/aap_gateway_api/oauth2_validator.py
+++ b/aap_gateway_api/oauth2_validator.py
@@ -1,5 +1,9 @@
+import logging
+
 from ansible_base.lib.workload_identity import SCOPE_REGISTRY
 from oauth2_provider.oauth2_validators import OAuth2Validator
+
+logger = logging.getLogger('aap_gateway_api.oauth2_validator')
 
 # Standard OIDC user identity claims advertised in discovery
 USER_IDENTITY_CLAIMS = ['sub', 'preferred_username', 'email', 'name', 'given_name', 'family_name']
@@ -45,9 +49,12 @@ class GatewayOIDCValidator(OAuth2Validator):
         Role claims use the aap_ prefix to avoid collision with other OIDC providers.
         Structured objects are used instead of flat strings because org/team names
         can contain arbitrary characters (including colons, slashes, etc.):
-        - aap_organizations: [{"name": "...", "roles": ["admin", "member"]}, ...]
-        - aap_teams: [{"name": "...", "organization": "...", "roles": ["admin", "member"]}, ...]
+        - aap_organizations: [{"name": "...", "roles": ["member"]}, ...]
+        - aap_teams: [{"name": "...", "organization": "...", "roles": ["member"]}, ...]
         - aap_system_role: scalar string (system_administrator, system_auditor, or normal_user)
+
+        The roles list contains all explicit assignments for a given object
+        (e.g. ["admin"], ["member"], or ["admin", "member"] when both are assigned).
 
         Claims reflect explicit RoleUserAssignment records only; implicit superuser
         access is not included (platform-level access is conveyed by aap_system_role).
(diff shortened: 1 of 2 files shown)


jay-steurer (approved):

Dostonbek1 (commented):
## 📋 Code Review Summary

This PR correctly identifies and fixes the root cause of OIDC group claims returning all orgs/teams for superusers: `access_qs()` short-circuits RBAC for users with `has_super_permission()`. The fix replaces `access_qs()` with direct `RoleUserAssignment` queries that only return explicit role assignments, which is the right approach. The `role` → `roles` format change is well-justified for representing users with multiple role assignments on the same object.

**Files Reviewed**: 2 files
**Comments**: 2 inline observations below

### 🔍 Issues Found
- 0 security/safety issues
- 1 correctness/robustness suggestion (defensive error handling)
- 1 code quality suggestion (deduplication)

### Overall Assessment
LGTM with minor suggestions. The fix is correct, the test coverage is comprehensive (superuser no-membership, superuser with explicit membership, dual roles, platform auditor), and the query optimization from 4 `get_or_create` calls to a single `filter` is a nice improvement.

### General Feedback
- Excellent PR description with root cause analysis, before/after examples, and query optimization details
- Good test structure that directly validates the bug-fix scenarios described in the PR
- Breaking change (`role` → `roles`) is properly documented with rationale

---

### `aap_gateway_api/oauth2_validator.py` — Lines 63-69 (role_defs lookup)

**Robustness**: If any of the four managed role definitions are missing from the database (e.g., during a partial migration or in a degraded state), the `role_defs[name]` dict access will raise an unhandled `KeyError`, resulting in a 500 on the userinfo endpoint.

While unlikely in production (these are created by DAB migrations), consider either:
1. Wrapping this in a guard that returns empty claims gracefully:
```python
role_defs = {
    rd.name: rd for rd in RoleDefinition.objects.filter(name__in=[...])
}
if len(role_defs) != 4:
    claims['aap_organizations'] = []
    claims['aap_teams'] = []
else:
    org_admin_id = role_defs['Organization Admin'].id
    # ... rest of logic ...
```
2. Or adding a brief comment explaining the assumption (if fail-fast is preferred here).

### `aap_gateway_api/oauth2_validator.py` — Lines 76-77, 89-90 (role accumulation)

**Nit (robustness)**: Using a `set` instead of a `list` for role accumulation would naturally deduplicate in the unlikely event of duplicate `RoleUserAssignment` records for the same user/object/role_definition:
```python
org_roles.setdefault(obj_id, set())
org_roles[obj_id].add('admin' if rd_id == org_admin_id else 'member')
```
`sorted()` already works with sets, so this is a drop-in change. Same applies to `team_roles`.

aap_gateway_api/oauth2_validator.py
```diff
@@ -45,31 +45,55 @@ def get_userinfo_claims(self, request):
...
             from aap_gateway_api.models import Organization, Team
 
-            # Determine org memberships with role type (admin vs member)
-            admin_org_ids = set(Organization.access_qs(user, 'change').values_list('pk', flat=True))
-            member_orgs = Organization.access_qs(user, 'member')
-            claims['aap_organizations'] = [{'name': org.name, 'role': 'admin' if org.pk in admin_org_ids else 'member'} for org in member_orgs]
+            role_defs = {
+                rd.name: rd for rd in RoleDefinition.objects.filter(name__in=['Organization Admin', 'Organization Member', 'Team Admin', 'Team Member'])
+            }
+            org_admin_id = role_defs['Organization Admin'].id
```
bhavenst: **Critical: Safety regression — unguarded `KeyError` if any `RoleDefinition` is missing**

The old code used `RoleDefinition.objects.managed.*` which calls `get_or_create` — self-healing and guaranteed to never fail. The new `filter()` + bare dict access will crash the **entire** OIDC userinfo endpoint with a `KeyError` if any of the four role definitions are missing (fresh install before migrations, corrupted DB, renamed managed role upstream in DAB).

This is security-critical: it takes down the auth endpoint for *all* users, not just the affected user.

Two options:

**Option A (recommended) — validate and degrade gracefully:**
```python
expected = {'Organization Admin', 'Organization Member', 'Team Admin', 'Team Member'}
missing = expected - role_defs.keys()
if missing:
    logger.error(
        "OIDC role claims unavailable: missing RoleDefinition(s): %s",
        ', '.join(sorted(missing)),
    )
    claims['aap_organizations'] = []
    claims['aap_teams'] = []
    # fall through to system_role below
else:
    # ... proceed with current logic
```

**Option B — use managed accessors for ID resolution only:**
```python
org_admin_id = RoleDefinition.objects.managed.org_admin.id
org_member_id = RoleDefinition.objects.managed.org_member.id
# ... etc
```
This preserves `get_or_create` safety while still using `RoleUserAssignment` queries for the actual claims (the core fix is orthogonal to how IDs are resolved).

Either way, a negative test (`RoleDefinition.objects.filter(...).delete()` then verify no crash) would lock in the safety guarantee.

aap_gateway_api/oauth2_validator.py
```diff
@@ -45,31 +45,55 @@ def get_userinfo_claims(self, request):
         Role claims use the aap_ prefix to avoid collision with other OIDC providers.
         Structured objects are used instead of flat strings because org/team names
         can contain arbitrary characters (including colons, slashes, etc.):
-        - aap_organizations: [{"name": "...", "role": "admin|member"}, ...]
-        - aap_teams: [{"name": "...", "organization": "...", "role": "admin|member"}, ...]
+        - aap_organizations: [{"name": "...", "roles": ["admin", "member"]}, ...]
```
bhavenst: **Suggestion: Docstring example is misleading**

Showing `"roles": ["admin", "member"]` implies every entry always contains both roles. In practice, most entries will have `["member"]` or `["admin"]` — the dual-role case is the exception.

Consider showing the common case:
```
- aap_organizations: [{"name": "...", "roles": ["member"]}, ...]
- aap_teams: [{"name": "...", "organization": "...", "roles": ["member"]}, ...]
```
With a note: *The `roles` list contains all explicit assignments (e.g., `["admin"]`, `["member"]`, or `["admin", "member"]` when both are assigned).*

The PR description's own "After" examples get this right — the docstring should match.

aap_gateway_api/oauth2_validator.py
```diff
@@ -45,31 +45,55 @@ def get_userinfo_claims(self, request):
...
+            org_roles = {}
+            for obj_id, rd_id in RoleUserAssignment.objects.filter(
+                user=user,
+                role_definition_id__in=[org_admin_id, org_member_id],
+            ).values_list('object_id', 'role_definition_id'):
+                org_roles.setdefault(obj_id, [])
+                org_roles[obj_id].append('admin' if rd_id == org_admin_id else 'member')
+
+            orgs_by_pk = {str(org.pk): org for org in Organization.objects.filter(pk__in=org_roles.keys())}
+            claims['aap_organizations'] = [
```
bhavenst: **Important: Non-deterministic claim ordering**

This iterates `org_roles.items()` whose order depends on DB query plan. Non-deterministic ordering can cause:
- Token cache misses in relying parties that compare tokens
- Debugging confusion (different tokens for same user look different)

Consider sorting the output:
```python
claims['aap_organizations'] = sorted(
    [{'name': orgs_by_pk[obj_id].name, 'roles': sorted(roles)}
     for obj_id, roles in org_roles.items() if obj_id in orgs_by_pk],
    key=lambda o: o['name'],
)
```
Same for `aap_teams` below.

aap_gateway_api/oauth2_validator.py
```diff
@@ -45,31 +45,55 @@ def get_userinfo_claims(self, request):
...
+                {'name': orgs_by_pk[obj_id].name, 'roles': sorted(roles)} for obj_id, roles in org_roles.items() if obj_id in orgs_by_pk
```
bhavenst: **Suggestion: Log orphaned role assignments instead of silently dropping them**

The `if obj_id in orgs_by_pk` guard correctly filters out assignments to deleted orgs — good defensive coding. But doing so silently means orphaned `RoleUserAssignment` records accumulate with no visibility to operators.

Consider:
```python
orphaned = set(org_roles.keys()) - set(orgs_by_pk.keys())
if orphaned:
    logger.warning(
        "User %s has RoleUserAssignment(s) for non-existent Organization(s): %s",
        user.pk, orphaned,
    )
```

bhavenst (changes requested):
## PR Review Summary

**Overall:** The core fix is correct — replacing `access_qs()` with `RoleUserAssignment` queries properly eliminates the superuser over-claim bug. The `"role"` → `"roles"` format change properly models dual admin+member. Test coverage for positive cases is excellent (superuser with/without assignments, dual roles, platform auditor). PR description is thorough.

However, there is one critical safety regression that should be addressed before merge, plus a couple of important improvements.

### Critical (1)
- **Unguarded `KeyError` on missing `RoleDefinition`** — see inline comment on `oauth2_validator.py` line 66

### Important (2)
- **Non-deterministic claim ordering** — see inline comment on line 80
- **No error handling around 5 DB queries in auth endpoint** — any DB error crashes the entire OIDC flow; wrapping the role claims section in a try-except that logs and returns partial claims would prevent a full outage

### Suggestions (3)
- **Docstring format example** — see inline comment on line 48
- **Silent filtering of orphaned assignments** — see inline comment on line 81
- **Missing test for orphaned `object_id`** — no test validates the `if obj_id in orgs_by_pk` guard; if someone refactors it away, claims would crash on deleted orgs/teams

### Strengths
- Core fix is correct and well-motivated
- Breaking change is justified and documented (feature is recent via AAP-68137)
- 4 new tests cover the key scenarios thoroughly
- Good query optimization (1 batch `filter()` vs 4 `get_or_create`)
- Removed comments that were just restating code logic — the new code is self-documenting via clear variable names

aap_gateway_api/oauth2_validator.py
zkayyali812: Implemented option A, will add a negative test case

aap_gateway_api/oauth2_validator.py
zkayyali812: docs string has been updated

aap_gateway_api/oauth2_validator.py
zkayyali812: **aap_organizations** is now sorted by name and **aap_teams** by (organization, name):

aap_gateway_api/oauth2_validator.py
zkayyali812: Added logging for both orphaned orgs/teams

bhavenst (approved):
All review findings addressed in `c407e5b`. Approving.

**Verified fixes:**
- **KeyError safety regression** — graceful degradation with `logger.error` and empty claims fallback ✅
- **Non-deterministic ordering** — `sorted()` by name for orgs, by `(organization, name)` for teams ✅
- **Docstring example** — shows common case `["member"]` with clarifying paragraph ✅
- **Orphan logging** — `logger.warning` for both orgs and teams ✅
- **Missing RoleDefinition test** — `test_missing_role_definitions_degrades_gracefully` locks in the safety guarantee ✅

**Bonus improvement:** Changed role accumulation from `list`/`append` to `set`/`add` — prevents duplicate entries if the DB has duplicate `RoleUserAssignment` records.

Clean, well-addressed update. LGTM.

PR #41, opened 2026-05-14
[AAP-75391]  Install ansible.platform Python runtime dependencies after collection install

rohitthakur2590:
## Description

- What is being changed?
The collection Makefile target now installs the ansible.platform collection's Python runtime dependencies after the collection is installed. If requirements.txt exists in the installed collection (2.7.20260514+), it installs from that file. If not (older releases), it falls back to installing requests directly.

- Why is this change needed?
ansible.platform 2.7.20260513 Galaxy release introduced a persistent manager subprocess (platform_manager.py) that runs outside Ansible's execution context and depends on the requests library for HTTP communication. Unlike the other connection modes which use ansible.module_utils.urls (available inside Ansible), this subprocess is a standalone Python process that requires requests to be explicitly installed.
The collection target was installing the collection but not its Python dependencies, causing make cleanup-services and any other target that runs playbooks using the collection to fail with:
ModuleNotFoundError: No module named 'requests'
This failure blocks all CI pipelines that use make cleanup-services after the 2.7.20260513 release landed on Galaxy.
- How does this change address the issue?
After ansible-galaxy collection install completes (via either the local or GitHub path), the target now checks for requirements.txt in the installed collection path and runs pip install -r against it. This is the canonical runtime dependency file declared by ansible.platform itself, so any future Python dependencies added to the collection will also be picked up automatically without further changes to aap-gateway.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [x] Development environment change
- [ ] Configuration change

## Self-Review Checklist

- [ ] I have added relevant comments to complex code sections
- [ ] I have updated documentation where needed
- [ ] I have considered the security impact of these changes
- [ ] I have considered performance implications
- [ ] I have thought about error handling and edge cases

### Steps to Test
1. 
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

## Summary by CodeRabbit

* **Chores**
  * Enhanced the build process to intelligently handle Python dependency installation, automatically detecting and installing required packages based on their availability.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,6 +178,7 @@ collection:
 			ansible-galaxy collection install git@github.com:ansible/ansible.platform.git; \
 		fi; \
 	fi;
+	pip install requests
 
 ## Build the docker containers
 docker-compose-build: tools/generated/sources update_django_ansible_base_hash tools/generated/.has_built_api


huffmanca:
Apologies, but I am slightly confused as to why the PR is being put here. The application installs `requests` in its requirements.in, and I'm wondering why the `requests` requirements in `ansible.platform/requirements` is insufficient? Why do we need a hardcoded install of `requests` here?

The unit test failure is unrelated to these changes, and it's non-blocking.

bhavenst (approved):
Hi, I don't see any harm in this but are you sure it will get requests where you want it (system python?) in the container?  I wonder if this should be done instead in https://github.com/ansible-automation-platform/automation-gateway-container

rohitthakur2590:
> https://github.com/ansible-automation-platform/automation-gateway-container

some how request not getting available in env , so I am trying this out, can we merge this so that I can check it this works , or else I'll revert it

huffmanca (approved):

PR #45, opened 2026-05-14
[AAP-74114] feat: add configurable concurrent session limit per user

bhavenst:
## Description

- **What:** Add a `SESSIONS_PER_USER` admin preference and `UserSessionMembership` model to limit concurrent authenticated sessions per user, with FIFO eviction of oldest sessions when the limit is exceeded.
- **Why:** Nothing currently prevents a single account from accumulating many live sessions, widening the blast radius of credential leaks and making it harder to reason about who is signed in. This was identified in spike [AAP-53067](https://redhat.atlassian.net/browse/AAP-53067).
- **How:** A `post_save` signal on Django's `Session` model tracks user-session mappings via a `UserSessionMembership` join table (OneToOne on Session). When a new session is created, memberships over the configured limit are evicted oldest-first using `SessionStore.delete()` to clear both the database and Redis cache. Expired sessions are excluded from the count. Eviction events are logged to the auth audit trail via `log_auth_event()`.

Semantics match AWX/Controller's `SESSIONS_PER_USER`:
- `-1` (default): unlimited, no enforcement
- `0`: only one session allowed (all others terminated on login)
- `N`: N additional concurrent sessions beyond the first (N+1 total)

## Intent Adherence: 95/100

- Solves the problem described in the story: 40/40
- No deviations from stated intent: 20/20
- No over-engineering or scope creep: 20/20
- A reviewer would agree the code matches the story: 15/20 (-5 additional integration tests in linked stories)

## Type of Change

## Self-Review Checklist

### Prerequisites
- Running gateway instance with PostgreSQL and Redis

### Steps to Test
1. Log in to the gateway admin and navigate to Settings > Configuration
2. Verify `SESSIONS_PER_USER` appears with default value `-1`
3. Set `SESSIONS_PER_USER` to `0`
4. Log in from two different browsers — the first session should be terminated when the second login occurs
5. Set `SESSIONS_PER_USER` to `1` — two concurrent sessions allowed, third login evicts the oldest
6. Set `SESSIONS_PER_USER` to `-1` — unlimited sessions, no eviction

### Expected Results
- Setting appears in the admin UI under Configuration
- Default `-1` allows unlimited sessions
- `0` allows only one session (FIFO eviction on each new login)
- Positive N allows N+1 total sessions
- Evicted sessions are immediately unusable (cleared from both DB and cache)
- Eviction events appear in the auth audit log

### Required Actions
- [x] Blocked by PR/MR: Integration tests covered by [AAP-40579](https://redhat.atlassian.net/browse/AAP-40579) and [AAP-40580](https://redhat.atlassian.net/browse/AAP-40580)

Assisted-by: Claude Code / Opus 4.6 (Anthropic)

## Summary by CodeRabbit

* **New Features**
  * Per-user session tracking with automatic termination of excess sessions.
  * New configurable preference to control maximum concurrent sessions per user (including an unlimited option).
* **Behavior Changes**
  * When limits apply, newest active sessions are kept; expired sessions are ignored and don’t count toward limits.
* **Tests**
  * Added comprehensive tests for tracking, duplicate prevention, eviction behavior (including cache removal and logging), cross-user isolation, and expired-session handling.

[AAP-53067]: https://redhat.atlassian.net/browse/AAP-53067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AAP-40579]: https://redhat.atlassian.net/browse/AAP-40579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/aap_gateway_api/models/user_session_membership.py
+++ b/aap_gateway_api/models/user_session_membership.py
@@ -38,7 +38,7 @@ class UserSessionMembership(models.Model):
         from aap_gateway_api.utils.preferences import get_setting
 
         try:
-            limit = get_setting('SESSIONS_PER_USER')
+            limit = get_setting('MAX_EXTRA_SESSIONS_PER_USER')
         except SettingNotSetException:
             return []
 
@@ -52,7 +52,7 @@ class UserSessionMembership(models.Model):
         # all memberships into Python.  select_for_update() prevents
         # two concurrent logins from both reading the same count and
         # each keeping their own session, exceeding the limit.
-        active = UserSessionMembership.objects.select_for_update().filter(user_id=user_id, session__expire_date__gt=now).order_by('-created')
+        active = UserSessionMembership.objects.select_for_update().filter(user_id=user_id, session__expire_date__gt=now).order_by('-created', '-pk')
 
         # limit is the number of *additional* sessions beyond the first.
         # So the total allowed is limit + 1.  A limit of 0 means only the
--- a/aap_gateway_api/models/user_session_membership.py
+++ b/aap_gateway_api/models/user_session_membership.py
@@ -1,0 +1,61 @@
+# Generated with AI assistance: Claude Code (Anthropic)
+import logging
+
+from django.conf import settings
+from django.contrib.sessions.models import Session
+from django.db import models
+from django.utils import timezone
+from django.utils.translation import gettext_lazy as _
+
+logger = logging.getLogger('aap.gateway.models.user_session_membership')
+
+
+class UserSessionMembership(models.Model):
+    user = models.ForeignKey(
+        settings.AUTH_USER_MODEL,
+        on_delete=models.CASCADE,
+        related_name='session_memberships',
+        help_text=_("The user who owns this session."),
+    )
+    session = models.OneToOneField(
+        Session,
+        on_delete=models.CASCADE,
+        related_name='membership',
+        help_text=_("The Django session associated with this membership."),
+    )
+    created = models.DateTimeField(default=timezone.now, help_text=_("When this session was first tracked."))
+
+    class Meta:
+        app_label = 'aap_gateway_api'
+
+    def __str__(self):
+        return f'{self.user_id} / {self.session_id}'
+
+    @staticmethod
+    def get_active_memberships_over_limit(user_id, now=None):
+        from ansible_base.lib.utils.settings import SettingNotSetException
+
+        from aap_gateway_api.utils.preferences import get_setting
+
+        try:
+            limit = get_setting('SESSIONS_PER_USER')
+        except SettingNotSetException:
+            return []
+
+        if limit == -1:
+            return []
+
+        if now is None:
+            now = timezone.now()
+
+        # Filter expired sessions at the DB level instead of loading
+        # all memberships into Python.  select_for_update() prevents
+        # two concurrent logins from both reading the same count and
+        # each keeping their own session, exceeding the limit.
+        active = UserSessionMembership.objects.select_for_update().filter(user_id=user_id, session__expire_date__gt=now).order_by('-created')
+
+        # limit is the number of *additional* sessions beyond the first.
+        # So the total allowed is limit + 1.  A limit of 0 means only the
+        # single newest session survives.
+        allowed = limit + 1
+        return list(active[allowed:])
--- a/aap_gateway_api/registered_preferences.py
+++ b/aap_gateway_api/registered_preferences.py
@@ -292,6 +292,23 @@ register(
 
 register(
     section="configuration",
+    preference_name="SESSIONS_PER_USER",
+    required=False,
+    default=-1,
+    preference_type="int_range",
+    help_text=_(
+        "Maximum number of simultaneous logged-in sessions allowed per user. "
+        "-1 means unlimited (no enforcement). "
+        "0 means only one session is allowed (any other session is terminated on login). "
+        "A positive number N allows N additional concurrent sessions beyond the first."
+    ),
+    min_value=-1,
+    encrypted=False,
+    label=_('Maximum Number of Simultaneous Logged In Sessions'),
+)
+
+register(
+    section="configuration",
     preference_name="DEFAULT_PAGE_SIZE",
     required=False,
     default=50,
--- a/aap_gateway_api/registered_preferences.py
+++ b/aap_gateway_api/registered_preferences.py
@@ -292,7 +292,7 @@ register(
 
 register(
     section="configuration",
-    preference_name="SESSIONS_PER_USER",
+    preference_name="MAX_EXTRA_SESSIONS_PER_USER",
     required=False,
     default=-1,
     preference_type="int_range",
@@ -303,6 +303,7 @@ register(
         "A positive number N allows N additional concurrent sessions beyond the first."
     ),
     min_value=-1,
+    max_value=100,
     encrypted=False,
     label=_('Maximum Number of Simultaneous Logged In Sessions'),
 )
--- a/aap_gateway_api/signals/session.py
+++ b/aap_gateway_api/signals/session.py
@@ -6,7 +6,7 @@ from ansible_base.lib.logging import log_auth_event
 from django.conf import settings
 from django.contrib.auth import SESSION_KEY, get_user_model
 from django.contrib.sessions.models import Session
-from django.db import IntegrityError
+from django.db import IntegrityError, transaction
 from django.db.models.signals import post_save
 from django.dispatch import receiver
 from django.utils import timezone
@@ -49,27 +49,24 @@ def track_user_session(sender, instance, **kwargs):
     if not user:
         return
 
-    # get_or_create handles the race where two concurrent requests
-    # both reach this point for the same session.  The OneToOneField
-    # on session enforces uniqueness at the DB level.
-    try:
-        _, created = UserSessionMembership.objects.get_or_create(
-            user_id=user_pk,
-            session=session,
-            defaults={'created': timezone.now()},
-        )
-    except IntegrityError:
-        # Another transaction already created the membership
-        return
+    with transaction.atomic():
+        try:
+            _, created = UserSessionMembership.objects.get_or_create(
+                user_id=user_pk,
+                session=session,
+                defaults={'created': timezone.now()},
+            )
+        except IntegrityError:
+            return
 
-    # Session updates (e.g. extending expiry) re-trigger post_save —
-    # only enforce the limit for genuinely new sessions.
-    if not created:
-        return
+        if not created:
+            return
 
-    expired_memberships = UserSessionMembership.get_active_memberships_over_limit(user_pk)
-    if expired_memberships:
-        session_keys = [m.session_id for m in expired_memberships]
-        for key in session_keys:
-            SessionStore(key).delete()
-        log_auth_event(f"Session limit enforced for user {user.username}: evicted {len(session_keys)} session(s) (SESSIONS_PER_USER limit exceeded)")
+        expired_memberships = UserSessionMembership.get_active_memberships_over_limit(user_pk)
+        if expired_memberships:
+            session_keys = [m.session_id for m in expired_memberships]
+            for key in session_keys:
+                SessionStore(key).delete()
+            log_auth_event(
+                f"Session limit enforced for user {user.username}: evicted {len(session_keys)} session(s) (MAX_EXTRA_SESSIONS_PER_USER limit exceeded)"
+            )
--- a/aap_gateway_api/signals/session.py
+++ b/aap_gateway_api/signals/session.py
@@ -1,0 +1,75 @@
+# Generated with AI assistance: Claude Code (Anthropic)
+import logging
+from importlib import import_module
+
+from ansible_base.lib.logging import log_auth_event
+from django.conf import settings
+from django.contrib.auth import SESSION_KEY, get_user_model
+from django.contrib.sessions.models import Session
+from django.db import IntegrityError
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from django.utils import timezone
+
+from aap_gateway_api.models.user_session_membership import UserSessionMembership
+
+logger = logging.getLogger('aap.gateway.signals.session')
+
+User = get_user_model()
+SessionStore = import_module(settings.SESSION_ENGINE).SessionStore
+
+
+@receiver(post_save, sender=Session)
+def track_user_session(sender, instance, **kwargs):
+    session = instance
+    try:
+        user_id = session.get_decoded().get(SESSION_KEY)
+    except Exception:
+        # Corrupted or tampered session data — Django's signing should
+        # prevent external tampering, but DB corruption or encoding
+        # errors can reach here.  Log so untracked sessions are visible.
+        logger.warning("Could not decode session; session is active but untracked by concurrent-session enforcement")
+        return
+
+    # Anonymous sessions (CSRF tokens, pre-login browsing) have no
+    # user in the session data — nothing to track or enforce.
+    if not user_id:
+        return
+
+    try:
+        user_pk = int(user_id)
+    except (ValueError, TypeError):
+        logger.warning("Non-numeric SESSION_KEY in session; skipping concurrent-session tracking")
+        return
+
+    # User may have been deleted between session creation and this
+    # signal (e.g. race with user removal).  No point tracking a
+    # session for a user that no longer exists.
+    user = User.objects.filter(pk=user_pk).first()
+    if not user:
+        return
+
+    # get_or_create handles the race where two concurrent requests
+    # both reach this point for the same session.  The OneToOneField
+    # on session enforces uniqueness at the DB level.
+    try:
+        _, created = UserSessionMembership.objects.get_or_create(
+            user_id=user_pk,
+            session=session,
+            defaults={'created': timezone.now()},
+        )
+    except IntegrityError:
+        # Another transaction already created the membership
+        return
+
+    # Session updates (e.g. extending expiry) re-trigger post_save —
+    # only enforce the limit for genuinely new sessions.
+    if not created:
+        return
+
+    expired_memberships = UserSessionMembership.get_active_memberships_over_limit(user_pk)
+    if expired_memberships:
+        session_keys = [m.session_id for m in expired_memberships]
+        for key in session_keys:
+            SessionStore(key).delete()
+        log_auth_event(f"Session limit enforced for user {user.username}: evicted {len(session_keys)} session(s) (SESSIONS_PER_USER limit exceeded)")
--- a/aap_gateway_api/tests/signals/test_session_signals.py
+++ b/aap_gateway_api/tests/signals/test_session_signals.py
@@ -43,8 +43,8 @@ class TestUserSessionMembership:
         store = SessionStore()
         store['some_key'] = 'some_value'
         store.create()
-        session = Session.objects.get(session_key=store.session_key)
-        with mock.patch.object(session, 'get_decoded', side_effect=ValueError("corrupted")):
+        with mock.patch.object(Session, 'get_decoded', side_effect=ValueError("corrupted")):
+            session = Session.objects.get(session_key=store.session_key)
             session.save()
         assert not UserSessionMembership.objects.filter(session_id=store.session_key).exists()
 
@@ -54,6 +54,18 @@ class TestUserSessionMembership:
         store[SESSION_KEY] = 'not-a-number'
         store.create()
         assert not UserSessionMembership.objects.filter(session_id=store.session_key).exists()
+
+    def test_no_membership_for_deleted_user(self):
+        """A session referencing a user PK that no longer exists should be skipped."""
+        store = SessionStore()
+        store[SESSION_KEY] = '999999'
+        store.create()
+        assert not UserSessionMembership.objects.filter(session_id=store.session_key).exists()
+
+    def test_str_representation(self, admin_user):
+        key = _create_session_for_user(admin_user)
+        membership = UserSessionMembership.objects.get(user=admin_user, session_id=key)
+        assert str(membership) == f'{admin_user.pk} / {key}'
 
     def test_session_update_does_not_trigger_eviction(self, admin_user, preference_manager):
         """Updating an existing session (e.g. extending expiry) must not create a new membership or trigger eviction."""
--- a/aap_gateway_api/tests/signals/test_session_signals.py
+++ b/aap_gateway_api/tests/signals/test_session_signals.py
@@ -38,6 +38,16 @@ class TestUserSessionMembership:
         store.create()
         assert not UserSessionMembership.objects.filter(session_id=store.session_key).exists()
 
+    def test_no_membership_for_corrupted_session(self):
+        """get_decoded() raising an exception should not crash the signal handler."""
+        store = SessionStore()
+        store['some_key'] = 'some_value'
+        store.create()
+        session = Session.objects.get(session_key=store.session_key)
+        with mock.patch.object(session, 'get_decoded', side_effect=ValueError("corrupted")):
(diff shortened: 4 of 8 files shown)


aap_gateway_api/registered_preferences.py
```diff
@@ -290,6 +290,22 @@
...
+    preference_type="int",
+    help_text=_(
+        "Maximum number of simultaneous logged-in sessions allowed per user. "
+        "-1 means unlimited (no enforcement). "
+        "0 means only one session is allowed (any other session is terminated on login). "
+        "A positive number N allows N additional concurrent sessions beyond the first."
+    ),
+    encrypted=False,
+    label=_('Maximum Number of Simultaneous Logged In Sessions'),
+)
```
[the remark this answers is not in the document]
bhavenst: Good point, fixed.

aap_gateway_api/signals/session.py
```diff
@@ -0,0 +1,48 @@
...
+        return
+
+    if not user_id:
+        return
+
+    if UserSessionMembership.objects.filter(user_id=user_id, session=session).exists():
+        return
+
+    if not User.objects.filter(pk=int(user_id)).exists():
+        return
```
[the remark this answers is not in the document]
bhavenst: Sure, we can do that.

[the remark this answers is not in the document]
bhavenst: Done!

john-westcott-iv:
## Code Review: [AAP-74114] feat: add configurable concurrent session limit per user

### Verdict: NEEDS_CHANGES

### Scores
| Lens | Score | Findings |
|---|---|---|
| Functionality | 7/10 | 3 Major |
| Security | 8.5/10 | 1 Major, 1 Minor, 1 Nit |
| Quality | 8.5/10 | 1 Major, 1 Minor |
| **Overall** | **8.0/10** | |

### Findings

#### Critical
(none)

#### Major

1. **[Functionality]** **[signals/session.py:exists/create]** TOCTOU race in membership creation. The `exists()` check and `create()` are not atomic -- under concurrent requests (two logins from the same user hitting different workers), both can pass `exists()` before either reaches `create()`. The `OneToOneField` on `session` will raise an unhandled `IntegrityError`, causing a 500 error. Note: AWX uses `ATOMIC_REQUESTS=True` which wraps every request in a database transaction, effectively serializing this pattern. Gateway does not have `ATOMIC_REQUESTS`, so the race condition is real here even though the same pattern was safe in AWX.
   - **Evidence**: `if UserSessionMembership.objects.filter(user_id=user_pk, session=session).exists(): return` followed by `UserSessionMembership.objects.create(...)` on separate lines with no atomicity guarantee.
   - **Confidence**: HIGH
   - **Fix**: Use `get_or_create()` which is atomic at the DB level:
     ```python
     _, created = UserSessionMembership.objects.get_or_create(
         user_id=user_pk, session=session,
         defaults={'created': timezone.now()},
     )
     if not created:
         return
     ```
   - **Points**: 1

2. **[Functionality]** **[signals/session.py:@receiver]** Wrong signal hook: `post_save` on `Session` fires on every `Session.save()`, not just login. Django's default `SESSION_SAVE_EVERY_REQUEST=True` means every authenticated request updates the session, re-triggering this handler. The handler ignores `kwargs['created']`, so on every request it queries the user, checks memberships, and evaluates the limit -- unnecessary work that also means lowering the limit mid-session triggers eviction on the next routine request (not just login).

   The `post_save` approach also forces decoding the session to identify the user and lacks access to the `request` object. The modern Django pattern is the **`user_logged_in` signal** (`django.contrib.auth.signals.user_logged_in`), which fires only on successful authentication and provides both `request` and `user` directly. This is the approach used by `django-single-session` and other current implementations. It would eliminate the session decode logic, the anonymous session guard, the non-numeric key guard, and the deleted-user check -- all of which exist only because `post_save` doesn't know who the user is.

   - **Evidence**: `@receiver(post_save, sender=Session)` and `def track_user_session(sender, instance, **kwargs)` -- the `created` kwarg from `post_save` is not used. No early return for updates.
   - **Confidence**: HIGH
   - **Fix** (minimum): Switch to `user_logged_in` signal:
     ```python
     from django.contrib.auth.signals import user_logged_in

     @receiver(user_logged_in)
     def track_user_session(sender, request, user, **kwargs):
         session_key = request.session.session_key
         # ... create membership and evict, with user already available
     ```
   - **Points**: 1

3. **[Functionality]** **[user_session_membership.py:get_active_memberships_over_limit]** Expired session filtering is done in Python instead of the database. The queryset loads ALL memberships into memory, then filters expired ones in a list comprehension. This defeats the database's ability to optimize the query and creates unbounded memory consumption for users with many historical sessions. Note: the in-memory filter only *skips* expired sessions from the count -- it does not delete them. Expired `UserSessionMembership` records accumulate indefinitely with no cleanup. The DB filter achieves the same filtering more efficiently without any cleanup tradeoff.
   - **Evidence**: `non_expired = [m for m in memberships if m.session.expire_date > now]` -- evaluates entire queryset into Python.
   - **Confidence**: HIGH
   - **Fix**: Push the filter to the database:
     ```python
     memberships = (
         UserSessionMembership.objects.select_related('session')
         .filter(user_id=user_id, session__expire_date__gt=now)
         .order_by('-created')
     )
     return list(memberships[allowed:])
     ```
   - **Points**: 1

4. **[Security]** **[signals/session.py:eviction]** Concurrent eviction race (double-spend). Two concurrent logins for the same user can both query `get_active_memberships_over_limit` before either has evicted any sessions. Both see the same "over-limit" set, attempt to delete the same sessions, but both keep their own newly-created session -- allowing 2 extra sessions to survive instead of the intended maximum.
   - **Evidence**: `UserSessionMembership.objects.create(...)` followed by `get_active_memberships_over_limit(user_pk)` followed by `SessionStore(key).delete()` -- no serialization between concurrent invocations for the same user.
   - **Confidence**: MEDIUM -- requires concurrent timing, but automated tooling or scripted attacks can achieve this reliably.
   - **Fix**: Wrap the create-check-evict sequence in `select_for_update()` on the user's memberships, or use a database advisory lock per user_id. This serializes the critical section per user without blocking other users.
   - **Points**: 1

#### Minor

1. **[Security]** **[signals/session.py:warning logs]** Session keys (bearer tokens) are logged at WARNING level in two places. If logs are forwarded to a centralized logging system, these could be harvested for session hijacking.
   - **Evidence**: `logger.warning(f"Could not decode session {session.session_key}; ...")` and `logger.warning(f"Non-numeric SESSION_KEY {user_id!r} in session {session.session_key}; ...")`
   - **Confidence**: MEDIUM -- depends on log access controls.
   - **Fix**: Truncate session keys: `session.session_key[:8] + '...'`
   - **Points**: 0.5

2. **[Security]** **[signals/session.py:eviction + user_session_membership.py]** The eviction code relies on `SessionStore.delete()` triggering ORM-level CASCADE deletes to clean up `UserSessionMembership` rows. We verified the gateway uses `ansible_base.lib.sessions.stores.cached_dynamic_timeout`, which extends Django's `cached_db` backend. The `cached_db.delete()` method deletes the DB row via `Session.objects.get().delete()` (triggering CASCADE) and also calls `self._cache.delete()` to clear Redis -- so both DB cleanup and cache clearing are handled correctly. However, this is an implicit assumption that could break if the session engine changes.
   - **Evidence**: `SessionStore(key).delete()` -- works correctly with current `cached_db`-based engine but depends on implementation details.
   - **Confidence**: LOW -- verified working with current engine, but fragile.
   - **Fix**: Add a code comment documenting the dependency: `# SessionStore.delete() must delete the django_session DB row (not just cache) for CASCADE cleanup of UserSessionMembership`.
   - **Points**: 0 (downgraded from 0.5 -- verified working with current engine)

3. **[Quality]** **[test_session_signals.py:class-level]** No test covers the behavior when an existing session is updated (not created). This is the most impactful missing test -- it would reveal Finding #2 (signal fires on every update).
   - **Evidence**: All tests create new sessions via `_create_session_for_user()`. No test calls `session.save()` on an existing session with an active membership and then checks whether eviction logic re-runs.
   - **Confidence**: HIGH
   - **Fix**: Add a test that creates a session, verifies membership, then updates the session (e.g., extends expiry), and asserts no duplicate membership and no eviction triggered.
   - **Points**: 1

4. **[Quality]** **[test_session_signals.py]** No test covers the concurrent creation race from Finding #1.
   - **Evidence**: No test calls `create()` twice for the same session or simulates concurrent signal handling.
   - **Confidence**: HIGH
   - **Fix**: Add a test that directly creates a duplicate membership and verifies the handler degrades gracefully (either via `get_or_create` returning existing, or `IntegrityError` being caught).
   - **Points**: 0.5

#### Nit

1. **[Functionality]** **[registered_preferences.py]** No `max_value` on `SESSIONS_PER_USER`. An admin could set it to 999999999, which is functionally equivalent to -1 but semantically confusing.
   - **Points**: 0

2. **[Quality]** **[signals/session.py:log_auth_event]** The `User` query (`User.objects.filter(pk=user_pk).first()`) exists primarily to get `user.username` for the audit log message. Consider logging `user_id={user_pk}` instead -- the PK is unambiguous (usernames can change) and eliminates the User query entirely. The existence check could then use the faster `.exists()` instead of `.first()`. (Note: if switching to `user_logged_in` signal per Finding #2, the `user` object is provided directly, making this moot.)
   - **Points**: 0

3. **[Security]** **[signals/session.py:eviction]** CASCADE cleanup of `UserSessionMembership` rows after eviction depends on `SessionStore.delete()` deleting the DB row (not just cache). Verified working with the current `cached_db`-based engine, but could break if the session engine changes. Add a documenting comment.
   - **Points**: 0

### Needs Human Judgment

- **Eviction semantics on limit change**: When an admin lowers `SESSIONS_PER_USER` from unlimited to 0, should existing sessions be evicted immediately (on next request) or only on next login? Finding #2's fix (check `created` kwarg) means eviction only happens on login. The current behavior (without the fix) evicts on next request. The team should decide which semantics they want.

### Verification Results
| min_value=-1 on preference | Visual inspection | `min_value=-1` present in register() call | PASS |
| int(user_id) guard | Visual inspection | try/except (ValueError, TypeError) present | PASS |
| FIFO eviction order | Logic trace | `order_by('-created')` + slice `[allowed:]` = oldest evicted | PASS |
| limit+1 semantics (0=one, N=N+1) | Logic trace | `allowed = limit + 1` matches AWX | PASS |
| AWX naming consistency | PR description | Matches AWX/Controller semantics | PASS |
| Anonymous session handling | Visual inspection | `if not user_id: return` | PASS |
| Deleted user handling | Visual inspection | `User.objects.filter(pk=user_pk).first()` with null check | PASS |
| session_memberships excluded from API | Visual inspection | Added to `ignore_relations` list | PASS |
| Session engine identified | `gh search code SESSION_ENGINE --repo ansible/jewel` | `ansible_base.lib.sessions.stores.cached_dynamic_timeout` | INFO |
| Engine extends cached_db | Read `cached_dynamic_timeout.py` source | `class SessionStore(CachedDBSessionStore)` -- inherits from Django's `cached_db` | PASS |
| SessionStore.delete() clears cache | Django `cached_db` source analysis | `delete()` calls `super().delete()` (DB row) + `self._cache.delete()` (Redis) | PASS |
| CASCADE cleanup on eviction | `cached_db.delete()` uses ORM `.delete()` on Session model | CASCADE to `UserSessionMembership` fires correctly | PASS |

### Observations

- The CodeRabbit findings (min_value validation and int() guard) are both already addressed in the code. CodeRabbit missed the existing `min_value=-1` parameter and the `try/except (ValueError, TypeError)` block.
- The FIFO eviction logic, limit+1 semantics, and AWX/Controller naming are all correct.
- The test suite covers the core happy paths well: membership creation, duplicate prevention, anonymous sessions, eviction at various limits, cache clearing, cross-user isolation, audit logging, and expired session handling.
- The model design (OneToOne on Session, FK on User, CASCADE on both) is correct.
- The `default=timezone.now` (callable reference) is correct in both model and migration.
- **Architecture note**: The `UserSessionMembership` join table exists because Django's default `Session` model has no FK to `User`. The package [django-qsessions](https://pypi.org/project/django-qsessions/) provides a drop-in session backend with a built-in `user` FK on sessions (plus `ip` and `user_agent`), which would eliminate the need for the join table and migration entirely. Adopting it would be a larger change (session backend swap) but worth considering as a follow-up to simplify this feature's architecture.

### Incidental Findings (out of scope -- not scored)

- No periodic cleanup mechanism for expired `UserSessionMembership` records. Over time, the table will accumulate rows for expired sessions. Consider adding a management command or periodic task for cleanup.
- The `log_auth_event()` call may lack request context when the signal fires outside a request (e.g., management command).

### Limitations

This review was performed by an AI agent in API-only mode. It does not understand business context, domain intent, organizational constraints, or deployment environment specifics. LOW-confidence findings and items in "Needs Human Judgment" require human verification.

**Partial review (API-only mode):** Codebase searches, lint verification, and reading files outside the diff were not available.

### Validation Summary
| Functionality | 4 | 0 | 0 | 3 Major, 1 Nit |
| Security | 7 | 2 (Nits), 1 (overlap), 1 (downgraded after engine verification) | 0 | 1 Major, 1 Minor, 1 Nit |
| Quality | 3 | 1 (verified OK) | 0 | 1 Major, 1 Minor |

### Path to 10/10
Each item is an agent-ready fix prompt:

1. **[Functionality] [signals/session.py]** (+1 point)
   > Replace the `exists()` + `create()` pattern with `get_or_create()`:
   > `_, created = UserSessionMembership.objects.get_or_create(user_id=user_pk, session=session, defaults={'created': timezone.now()})`
   > Return early if `not created`.

2. **[Functionality] [signals/session.py]** (+1 point)
   > Switch from `@receiver(post_save, sender=Session)` to `@receiver(user_logged_in)`.
   > The handler receives `request` and `user` directly, eliminating session decode,
   > anonymous session guard, non-numeric key guard, and deleted-user check.
   > Fires only on login, not on every session update.

3. **[Functionality] [user_session_membership.py]** (+1 point)
   > Push expired-session filtering to the database:
   > `.filter(user_id=user_id, session__expire_date__gt=now)` instead of Python list comprehension.
   > Use `list(memberships[allowed:])` to push slicing to DB too.

4. **[Security] [signals/session.py]** (+1 point)
   > Serialize the create-check-evict sequence per user using `select_for_update()`
   > or a database advisory lock to prevent the concurrent eviction race.

### Final Score
**Overall: 8.0/10 -- NEEDS_CHANGES**

[AAP-74114]: https://redhat.atlassian.net/browse/AAP-74114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

bhavenst:
@john-westcott-iv 

Race condition: Agreed, fixed

The SESSION_SAVE_EVERY_REQUEST issue is definitely interesting.  The review is wrong though.  The default is False and we don't change that, so this doesn't really apply.

SESSION_SAVE_EVERY_REQUEST	False

Also fixing db filtering vs python, concurrent login bug, adding UT and truncate session key.  The rest I don't feel like are needed.

john-westcott-iv (changes requested):
## Re-review: Concurrent Session Limit

The two prior CodeRabbit concerns are both addressed (`int_range` with `min_value=-1`, guarded `int()` cast). The code also has nice improvements since last pass — `get_or_create` for atomicity, `select_for_update()`, DB-level expiry filtering.

One issue to address before approval:

### `select_for_update()` is ineffective without `transaction.atomic()`

Gateway does not set `ATOMIC_REQUESTS` in its database configuration (confirmed — zero references in the repo). This means Django runs in **autocommit mode**, where each query executes in its own implicit transaction. The `select_for_update()` in `get_active_memberships_over_limit` acquires a row lock that is immediately released when the implicit transaction ends after that single query.

The race scenario: two concurrent logins both call `get_or_create` (each succeeds — different sessions), then both call `get_active_memberships_over_limit`. Without an explicit transaction boundary, both reads see the same membership count, both decide their session is within the limit, and the limit is exceeded.

This pattern worked in AWX/Controller because AWX uses `ATOMIC_REQUESTS = True`, so every request is wrapped in a transaction automatically. Gateway doesn't have that.

**Fix:** Wrap the create + check + evict logic in `transaction.atomic()`:

```python
from django.db import transaction

@receiver(post_save, sender=Session)
def track_user_session(sender, instance, **kwargs):
    # ... decode session, validate user_id, check user exists ...

    with transaction.atomic():
        try:
            _, created = UserSessionMembership.objects.get_or_create(
                user_id=user_pk,
                session=session,
                defaults={'created': timezone.now()},
            )
        except IntegrityError:
            return

        if not created:
            return

        expired_memberships = UserSessionMembership.get_active_memberships_over_limit(user_pk)
        if expired_memberships:
            session_keys = [m.session_id for m in expired_memberships]
            for key in session_keys:
                SessionStore(key).delete()
            log_auth_event(...)
```

This ensures `select_for_update()` holds the lock until after eviction completes, preventing concurrent logins from both reading stale counts.

aap_gateway_api/tests/signals/test_session_signals.py
```diff
@@ -0,0 +1,189 @@
+# Generated with AI assistance: Claude Code (Anthropic)
```
huffmanca, line 1: I see this comment in a couple of files; do we need to start including it?

aap_gateway_api/registered_preferences.py
```diff
@@ -290,6 +290,23 @@
...
+    section="configuration",
+    preference_name="SESSIONS_PER_USER",
+    required=False,
+    default=-1,
+    preference_type="int_range",
+    help_text=_(
+        "Maximum number of simultaneous logged-in sessions allowed per user. "
+        "-1 means unlimited (no enforcement). "
+        "0 means only one session is allowed (any other session is terminated on login). "
+        "A positive number N allows N additional concurrent sessions beyond the first."
```
huffmanca, line 303: I understand that this is to have parity with Controller's prior implementation, but it still feels obscure based on the name, since this is really controlling an additional number of sessions per user. 

Do we want to take this opportunity to be clearer with either the preference name or the functionality? Or is it more important to have strict adherence to 2.4 functionality?

aap_gateway_api/signals/session.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Session updates (e.g. extending expiry) re-trigger post_save —
+    # only enforce the limit for genuinely new sessions.
+    if not created:
+        return
+
+    expired_memberships = UserSessionMembership.get_active_memberships_over_limit(user_pk)
+    if expired_memberships:
+        session_keys = [m.session_id for m in expired_memberships]
```
Dostonbek1: **[Functionality] Major — Burst eviction under limit reduction**

When `SESSIONS_PER_USER` is changed from unlimited to a low value (e.g., 0) on a system where a user has many active sessions, a single login will cascade-delete all excess sessions within the `SessionStore.save()` transaction, holding `select_for_update` row locks for the entire duration. This could cause concurrent login attempts for the same user to block.

Consider documenting this operational consideration, or capping eviction per signal invocation (e.g., evict at most N sessions at a time).

---

**Note on the existing `transaction.atomic()` concern:** Django's `SessionStore.save()` in `backends/db.py:126` already wraps `obj.save()` in `transaction.atomic()`, and `post_save` fires inside that block — so `select_for_update()` IS effective here. The suggestion to add an explicit `transaction.atomic()` wrapper would create a nested savepoint, which is harmless but unnecessary.

aap_gateway_api/registered_preferences.py
```diff
@@ -290,6 +290,23 @@
...
+    ),
+    min_value=-1,
+    encrypted=False,
+    label=_('Maximum Number of Simultaneous Logged In Sessions'),
+)
```
Dostonbek1, line 309: **[Quality] Minor — Missing `max_value` on `SESSIONS_PER_USER`**

Every other `int_range` preference in this file has an explicit `max_value`. Without one, `IntRangePreference` defaults to `DEFAULT_MAX_VALUE = 100`, silently capping `SESSIONS_PER_USER` at 100 with no mention in `help_text`.

Consider adding an explicit `max_value` for consistency and clarity:
```suggestion
    min_value=-1,
    max_value=100,
    encrypted=False,
```

aap_gateway_api/signals/session.py
```diff
@@ -0,0 +1,75 @@
...
+
+@receiver(post_save, sender=Session)
+def track_user_session(sender, instance, **kwargs):
+    session = instance
+    try:
+        user_id = session.get_decoded().get(SESSION_KEY)
+    except Exception:
+        # Corrupted or tampered session data — Django's signing should
+        # prevent external tampering, but DB corruption or encoding
+        # errors can reach here.  Log so untracked sessions are visible.
```
Dostonbek1, line 30: **[Quality] Minor — Missing test coverage for corrupted session path**

This `except Exception` guard on `get_decoded()` is a reasonable defensive path, but it has no test coverage. A test with a mocked `get_decoded` raising `ValueError` would verify the handler returns gracefully without creating a membership.

aap_gateway_api/models/user_session_membership.py
```diff
@@ -0,0 +1,61 @@
...
+            return []
+
+        if now is None:
+            now = timezone.now()
+
+        # Filter expired sessions at the DB level instead of loading
+        # all memberships into Python.  select_for_update() prevents
+        # two concurrent logins from both reading the same count and
+        # each keeping their own session, exceeding the limit.
+        active = UserSessionMembership.objects.select_for_update().filter(user_id=user_id, session__expire_date__gt=now).order_by('-created')
```
Dostonbek1: **[Quality] Nit — Ordering tie-break is undefined**

If two memberships have identical `created` timestamps, eviction order is non-deterministic. Adding `-pk` as a tiebreaker would guarantee deterministic FIFO:
```suggestion
        active = UserSessionMembership.objects.select_for_update().filter(user_id=user_id, session__expire_date__gt=now).order_by('-created', '-pk')
```

Dostonbek1 (commented):
Review: 9.3/10 — 1 Major, 2 Minor, 1 Nit. Overall solid implementation with correct signal registration, proper CASCADE behavior, and effective select_for_update() within Django's existing session transaction. See inline comments for details.

aap_gateway_api/tests/signals/test_session_signals.py
bhavenst: As far as I understand, this one is optional.  It was added by carbonite so I just left it..

aap_gateway_api/registered_preferences.py
bhavenst: Yeah I had the same thought on the setting name but went back and saw that it's the same as awx.  Even so though I checked with Ron and he's OK with MAX_EXTRA_SESSIONS_PER_USER, which I think is better.  The implementation stays though.

aap_gateway_api/signals/session.py
bhavenst: This matches Controller's existing behavior — when the limit is reduced, excess sessions are evicted on the next login. Since reducing the limit is a deliberate admin action (not something that happens automatically), the burst eviction is expected and intentional. Capping eviction per invocation would leave the system in a state where the limit isn't actually enforced until enough logins occur, which would be more confusing operationally. We'll keep this as-is for Controller parity.

john-westcott-iv (approved):
Re-review: our prior concern (select_for_update without transaction.atomic) is fully addressed. The create + check + evict flow is now wrapped in transaction.atomic(), which ensures the row lock from select_for_update is held through eviction. Other reviewer feedback (setting rename, max_value, corrupted session test, ordering tiebreak) also addressed. LGTM.

PR #48, opened 2026-05-15
[AAP-61226] Update tests: replace removed FEATURE_CASE_INSENSITIVE_AUTH_MAPS_ENABLED flag

Dostonbek1:
## Summary

Companion PR to [ansible/django-ansible-base#1004](https://github.com/ansible/django-ansible-base/pull/1004), which removes the `FEATURE_CASE_INSENSITIVE_AUTH_MAPS_ENABLED` feature flag from DAB (making case-insensitive auth maps permanent behavior).

This PR updates gateway feature flag tests that referenced the now-removed flag.

**Why:** With the Runtime Feature Flag UI shipping in 2.7 (and backported to 2.6), customers would see this Developer Preview flag. Since the behavior is now permanent, the flag has been removed from DAB and these tests need to reference a flag that still exists.

## Changes

- **test_feature_flag_api.py**: Replace all references to `FEATURE_CASE_INSENSITIVE_AUTH_MAPS_ENABLED` with `FEATURE_INDIRECT_NODE_COUNTING_ENABLED` (the only remaining runtime-toggleable flag). Adjust install-time precedence test assertions to match the new flag's default value (`False` instead of `True`).

## Dependencies

> **This PR depends on [ansible/django-ansible-base#1004](https://github.com/ansible/django-ansible-base/pull/1004)** — the DAB PR must be merged first so that the flag definition is removed and the `django-ansible-base` pin is updated before this PR's tests will pass in CI.

## Test plan

- [x] All 25 feature flag tests pass locally (`tox -e py312` scoped to `aap_gateway_api/tests/feature_flags`)
- [ ] CI passes after DAB PR is merged and pin is bumped

## Jira

AAP-61226

### DIFF
--- a/aap_gateway_api/tests/feature_flags/test_feature_flag_api.py
+++ b/aap_gateway_api/tests/feature_flags/test_feature_flag_api.py
@@ -322,7 +322,7 @@ def test_feature_flags_detail_patch_locked_by_settings(admin_api_client, runtime
     - If a feature flag is set at install time, it becomes READ-ONLY and cannot be changed at runtime
     - Runtime feature flags can only be toggled if they were NOT explicitly set at install time
     """
-    feature_flag = "FEATURE_CASE_INSENSITIVE_AUTH_MAPS_ENABLED"
+    feature_flag = "FEATURE_INDIRECT_NODE_COUNTING_ENABLED"
     try:
         created_flag = AAPFlag.objects.get(name=feature_flag)
     except AAPFlag.DoesNotExist:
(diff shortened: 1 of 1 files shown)


john-westcott-iv (approved):
## Code Review: Update tests — replace removed FEATURE_CASE_INSENSITIVE_AUTH_MAPS_ENABLED flag

**Verdict**: LGTM

Clean companion to django-ansible-base#1004. All 22 references to `FEATURE_CASE_INSENSITIVE_AUTH_MAPS_ENABLED` are replaced with `FEATURE_INDIRECT_NODE_COUNTING_ENABLED`. The install-time precedence tests correctly invert assertion values and `override_settings` values to account for the new flag's `False` default (vs the old flag's `True` default) — the test logic is preserved, just exercised from the opposite direction.

PR #86, opened 2026-05-29
AAP-65883: Add gateway cache invalidation task

bhavenst:
## Summary

- Add `clear_gateway_cache` dispatcherd task wrapping DAB's `clear_cache()` on the `gateway_broadcast` queue
- Register the task in `pre_fork.py` for worker discovery
- 4 unit tests covering delegation, key deletion, empty list, and module importability

## Dependencies

- Builds upon ansible/django-ansible-base#1014

## Test plan

- [x] `ruff check` and `ruff format --check` pass on all changed files
- [x] Unit tests in `aap_gateway_api/tests/tasks/test_cache.py`
- [ ] CI passes

Jira: https://redhat.atlassian.net/browse/AAP-65883

Assisted-by: Claude Code / Opus 4.6 (Anthropic)

## Summary by CodeRabbit

* **New Features**
  * Added a task to clear gateway cache; it removes only specified keys and is a no-op for an empty list.

* **Tests**
  * Added tests for targeted key deletion, delegation behavior, empty-list no-op, and module importability.

* **Chores**
  * Updated coverage/analysis exclusions and adjusted startup/shutdown initialization ordering to ensure clean cache and connection teardown.

### DIFF
--- a/aap_gateway_api/dispatch/pre_fork.py
+++ b/aap_gateway_api/dispatch/pre_fork.py
@@ -4,7 +4,7 @@ from django.db import connections
 
 django.setup()
 
-import aap_gateway_api.tasks.cache  # noqa: F401, E402
+import aap_gateway_api.tasks.cache  # noqa: F401, E402 — register dispatcherd task
 
 cache.close()
 connections.close_all()
--- a/aap_gateway_api/dispatch/pre_fork.py
+++ b/aap_gateway_api/dispatch/pre_fork.py
@@ -4,5 +4,7 @@ from django.db import connections
 
 django.setup()
 
+import aap_gateway_api.tasks.cache  # noqa: F401, E402
+
 cache.close()
 connections.close_all()
--- a/aap_gateway_api/tasks/cache.py
+++ b/aap_gateway_api/tasks/cache.py
@@ -1,0 +1,7 @@
+from ansible_base.lib.cache.tasks import clear_cache
+from dispatcherd.publish import task
+
+
+@task(queue='gateway_broadcast', timeout=600)
+def clear_gateway_cache(cache_keys):
+    clear_cache(cache_keys)
--- a/aap_gateway_api/tasks/cache.py
+++ b/aap_gateway_api/tasks/cache.py
@@ -3,5 +3,5 @@ from dispatcherd.publish import task
 
 
 @task(queue='gateway_broadcast', timeout=600)
-def clear_gateway_cache(cache_keys):
+def clear_gateway_cache(cache_keys: list[str]) -> None:
     clear_cache(cache_keys)
--- a/aap_gateway_api/tests/tasks/test_cache.py
+++ b/aap_gateway_api/tests/tasks/test_cache.py
@@ -29,7 +29,7 @@ def test_clear_gateway_cache_deletes_keys(cache_backend):
     assert cache_backend.get('gw_key_b') == 'val_b'
 
 
-def test_clear_gateway_cache_delegates_to_dab(cache_backend):
+def test_clear_gateway_cache_delegates_to_dab():
     with patch('aap_gateway_api.tasks.cache.clear_cache') as mock_clear:
         from aap_gateway_api.tasks.cache import clear_gateway_cache
 
--- a/aap_gateway_api/tests/tasks/test_cache.py
+++ b/aap_gateway_api/tests/tasks/test_cache.py
@@ -1,0 +1,53 @@
+from unittest.mock import patch
+
+import pytest
+
+
+@pytest.fixture
+def cache_backend(settings):
+    settings.CACHES = {
+        'default': {
+            'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
+            'LOCATION': 'test-gateway-cache',
+        }
+    }
+    from django.core.cache import cache
+
+    cache.clear()
+    cache.set('gw_key_a', 'val_a')
+    cache.set('gw_key_b', 'val_b')
+    yield cache
+    cache.clear()
+
+
+def test_clear_gateway_cache_deletes_keys(cache_backend):
+    from aap_gateway_api.tasks.cache import clear_gateway_cache
+
+    clear_gateway_cache(['gw_key_a'])
+
+    assert cache_backend.get('gw_key_a') is None
+    assert cache_backend.get('gw_key_b') == 'val_b'
+
+
+def test_clear_gateway_cache_delegates_to_dab(cache_backend):
+    with patch('aap_gateway_api.tasks.cache.clear_cache') as mock_clear:
+        from aap_gateway_api.tasks.cache import clear_gateway_cache
+
+        clear_gateway_cache(['key_x', 'key_y'])
+
+        mock_clear.assert_called_once_with(['key_x', 'key_y'])
+
+
+def test_clear_gateway_cache_empty_list(cache_backend):
+    from aap_gateway_api.tasks.cache import clear_gateway_cache
+
+    clear_gateway_cache([])
+
+    assert cache_backend.get('gw_key_a') == 'val_a'
+    assert cache_backend.get('gw_key_b') == 'val_b'
+
+
+def test_tasks_cache_module_importable():
+    import aap_gateway_api.tasks.cache as cache_module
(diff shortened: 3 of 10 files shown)


aap_gateway_api/tasks/cache.py
```diff
@@ -0,0 +1,7 @@
+from ansible_base.lib.cache.tasks import clear_cache
+from dispatcherd.publish import task
+
+
+@task(queue='gateway_broadcast', timeout=600)
```
jay-steurer, line 5: **Nit:** `timeout=600` (10 minutes) seems generous for cache key deletion, which should be near-instantaneous even for large key lists. Was this chosen to match an existing convention, or would a lower value (e.g. 30–60s) be more appropriate? A long timeout on a broadcast queue could mask a hung worker.

aap_gateway_api/tasks/cache.py
```diff
@@ -0,0 +1,7 @@
...
+def clear_gateway_cache(cache_keys):
```
jay-steurer: **Suggestion:** Consider adding a type annotation for `cache_keys` — e.g. `cache_keys: list[str]` — so callers know the expected shape without having to trace into DAB's `clear_cache`.

aap_gateway_api/tests/tasks/test_cache.py
```diff
@@ -0,0 +1,53 @@
...
+    from aap_gateway_api.tasks.cache import clear_gateway_cache
+
+    clear_gateway_cache(['gw_key_a'])
+
+    assert cache_backend.get('gw_key_a') is None
+    assert cache_backend.get('gw_key_b') == 'val_b'
+
+
+def test_clear_gateway_cache_delegates_to_dab(cache_backend):
+    with patch('aap_gateway_api.tasks.cache.clear_cache') as mock_clear:
```
jay-steurer, line 33: **Bug:** This test takes the `cache_backend` fixture but doesn't need it — the `patch` mock intercepts `clear_cache` before it touches the real cache. The fixture adds unnecessary setup/teardown. Removing it would make the test's intent clearer ("we only care about delegation, not actual cache state").

```python
def test_clear_gateway_cache_delegates_to_dab():
```

aap_gateway_api/dispatch/pre_fork.py
```diff
@@ -4,5 +4,7 @@
 
 django.setup()
 
+import aap_gateway_api.tasks.cache  # noqa: F401, E402
```
jay-steurer: **Nit (readability):** This file already binds `cache` from `django.core.cache`. The full-path import doesn't shadow it, but a reader scanning quickly might wonder. A trailing comment like `# register dispatcherd task` (instead of or in addition to the noqa) would signal intent more clearly than the linter suppression alone.

jay-steurer (commented):
Good, clean PR — thin wrapper + focused tests. A few items below worth a look.

john-westcott-iv (commented):
## Code Review: AAP-65883 — Gateway cache invalidation task

**Score**: 9.5/10 (Functionality 10, Security 10, Quality 9)

Clean thin wrapper with good tests. One architectural observation:

### Observation

**No callers yet** — `clear_gateway_cache` is defined and registered for dispatcherd discovery, but nothing in the Gateway codebase currently calls `.delay()` on it. This means the task exists but is never triggered. Presumably the wiring (e.g., in a view's `perform_update` or a signal handler via `connection.on_commit`) is planned for a follow-up.

If the intent is on-demand-only invalidation (no periodic sweep), that's the right design — just confirming the invocation path is planned for a subsequent PR.

### Minor

The `timeout=600` on the task decorator is the dispatcherd execution timeout (max time a single invocation can run), not a polling interval. This is fine for a cache invalidation that should complete in milliseconds.

No blocking issues. CI is green (codecov failures are reporting-only).

aap_gateway_api/tasks/cache.py
bhavenst: Kept `timeout=600` to match the AWX convention (`tower_settings_change` uses the same value). John also confirmed this is fine — it's the dispatcherd execution timeout, and cache deletion is near-instantaneous, so it's effectively just a safety ceiling.

aap_gateway_api/tasks/cache.py
bhavenst: Added in ffc5cca7 — `cache_keys: list[str]` and `-> None` return type.

aap_gateway_api/tests/tasks/test_cache.py
bhavenst: Good catch — removed the unnecessary fixture in ffc5cca7.

aap_gateway_api/dispatch/pre_fork.py
bhavenst: Added `— register dispatcherd task` comment in ffc5cca7.

bhavenst:
Thanks @jay-steurer and @john-westcott-iv for the reviews. All addressed in ffc5cca7:

- **Type annotation**: Added `cache_keys: list[str]` and `-> None` return type
- **Test fixture**: Removed unnecessary `cache_backend` from delegation test
- **Pre-fork comment**: Added `— register dispatcherd task` for clarity
- **timeout=600**: Kept to match AWX convention (safety ceiling, not a performance concern)

Re: **no callers yet** (John) — correct, invocation wiring (e.g., from Gateway's `preference_updated` signal or views via `connection.on_commit()`) is planned for a follow-up story.

Rebased to latest devel.

jay-steurer (commented):
## Code Review: AAP-65883 — Gateway cache invalidation task (Second Review)

### First-Review Feedback Resolution

All three actionable items from the first review were addressed in `ffc5cca7`:

| Feedback | Status |
|---|---|
| Add type annotations to `clear_gateway_cache` | Fixed — `cache_keys: list[str]` and `-> None` added |
| Remove unnecessary `cache_backend` fixture from delegation test | Fixed — fixture parameter removed |
| Add descriptive comment on `pre_fork.py` import | Fixed — `— register dispatcherd task` appended to noqa |

All cleanly done. No over-corrections.

### Second-Review Findings

**The code itself is solid.** Thin wrapper, correct queue choice, sensible tests. A few observations on what remains:

**1. CI status — "fix-ci" label is still present**

- Konflux ATF tests: **FAIL** (15m)
- codecov/patch + codecov/project: **FAIL** (reporting-only, non-blocking)
- All lint, unit, service, and view tests: **PASS**

The ATF failure needs investigation before removing the `fix-ci` label. If the failure is unrelated to this PR (e.g., infrastructure flakiness or a pre-existing issue in devel), it should be documented in the PR thread so reviewers know it's not a regression.

**2. "blocked" label — dependency on DAB #1014**

The PR correctly declares its dependency on `ansible/django-ansible-base#1014` (`ansible_base.lib.cache.tasks.clear_cache`). The import will fail at worker startup if DAB is merged without that PR. This is the right sequencing — just confirming the label should stay until DAB #1014 merges.

**3. No new issues found in the code**

- `gateway_broadcast` queue is already configured in `dispatch/config.py:24` — the task will be routed correctly
- The `pre_fork.py` import placement (after `django.setup()`, before `cache.close()`) is correct for dispatcherd task registration
- No naming collision between `from django.core.cache import cache` and `import aap_gateway_api.tasks.cache` — they bind different names (`cache` vs `aap_gateway_api`)
- The `LocMemCache` fixture with unique `LOCATION` avoids cross-test contamination
- This is the first `@task()` decorator in the gateway codebase — establishes the pattern for future dispatcherd tasks under `aap_gateway_api/tasks/`

### Verdict

**LGTM** — Code is clean, first-review feedback fully addressed, tests are focused and correct. The two blockers are external: DAB dependency and the ATF test failure. Once those are resolved and labels removed, this is ready to merge.

**Score: 9.5/10** (same as first review — nothing degraded, fixes were mechanical and correct)

jay-steurer (approved):

john-westcott-iv (approved):
## Re-review: Feedback Addressed

- Type annotation added to `clear_gateway_cache(cache_keys: list[str]) -> None`
- Pre-fork import comment clarified

All functional CI checks passing (codecov reporting failures are non-blocking). Approved.

PR #95, opened 2026-06-01
[AAP-59893] feat: add sidecar Redis to Gateway dev environment

jeffh-oss:
## Summary

- Add a sidecar Redis container (Unix socket, no TLS/ACL) as the `"default"` cache backend for the Gateway dev environment
- Existing centralized Redis remains under `"legacy"` and `"primary"` cache entries
- Cluster/sentinel Redis modes are untouched

## Changes

| File | Change |
|------|--------|
| `redis-sidecar.conf.j2` (new) | Minimal Redis config: Unix socket only, no TCP/TLS/ACL, db=4 |
| `docker-compose.yml.j2` | Sidecar service (inside gateway loop), depends_on/volume, named volumes |
| `docker-compose-stage.yml.j2` | Same sidecar changes for staging template |
| `initialize_redis.yml` | Add sidecar template to Ansible render loop |
| `defaults.py` | Sidecar as `"default"` cache, old default renamed to `"legacy"` |
| `Makefile` | Add sidecar template as build dependency |
| `test_settings_utils.py` | Tests for sidecar as default cache + legacy entry |

## Verification Steps

After `make docker-compose` with all 5 containers up (db, redis, redis_sidecar, gateway, proxy):

### Step 1: Verify sidecar Redis responds via Unix socket
```bash
podman exec aap_gw_redis_sidecar_1 redis-cli -s /var/run/redis/redis.sock ping
```
Expected: `PONG`

### Step 2: Verify the Gateway can see the socket
```bash
podman exec aap_gw_1 ls -la /var/run/redis/redis.sock
```

### Step 3: Verify the Gateway is using the sidecar as the default cache
```bash
podman exec aap_gw_1 python3 -c "
from django.conf import settings
print('Default cache LOCATION:', settings.CACHES['default']['LOCATION'])
print('Client class:', settings.CACHES['default']['OPTIONS']['CLIENT_CLASS'])
print('Legacy entry:', settings.CACHES['legacy']['BACKEND'])
"
```
Expected:
- `Default cache LOCATION: unix:///var/run/redis/redis.sock?db=4`
- `Client class: django_redis.client.DefaultClient`
- `Legacy entry: ansible_base.lib.cache.fallback_cache.DABCacheWithFallback`

### Step 4: Write and read a test key through the default cache
```bash
podman exec aap_gw_1 python3 -c "
from django.core.cache import cache
cache.set('sidecar_test', 'hello from sidecar', timeout=60)
print('Read back:', cache.get('sidecar_test'))
"
```
Expected: `Read back: hello from sidecar`

### Step 5: Confirm the key landed in the sidecar (not centralized Redis)
```bash
podman exec aap_gw_redis_sidecar_1 redis-cli -s /var/run/redis/redis.sock -n 4 keys "*"
```
Should show a key like `gateway:sidecar_test`. Then verify it is NOT in the centralized Redis:
```bash
podman exec aap_gw_redis_1 redis-cli -p 6380 --tls --cert /usr/local/etc/redis/redis.cert --key /usr/local/etc/redis/redis.key --cacert /usr/local/etc/redis/ca.cert --user gateway1 --pass redis_password keys "*"
```

### Step 6: Verify the centralized Redis (legacy/primary) still works
```bash
podman exec aap_gw_1 python3 -c "
from django.core.cache import caches
primary = caches['primary']
primary.set('legacy_test', 'hello from centralized redis', timeout=60)
print('Read back:', primary.get('legacy_test'))
"
```
Expected: `Read back: hello from centralized redis`

## Test plan

- [x] `make docker-compose` starts with all 5 containers
- [x] Sidecar Redis responds to PING via Unix socket
- [x] Gateway uses sidecar as default cache
- [x] Cache read/write works through sidecar
- [x] Keys land in sidecar Redis, not centralized Redis
- [ ] Centralized Redis (primary) still works for read/write
- [x] Unit tests pass (ruff lint + format clean)
- [ ] Existing sentinel/cluster modes unaffected

## Summary by CodeRabbit

* **Chores**
  * Default cache now uses a Unix-socket Redis sidecar for local connections.
  * Per-gateway Redis sidecar instances added to deployments, with corresponding startup ordering and socket-backed volumes.
  * Health checks updated to detect and use Unix-socket Redis endpoints.
  * Build/regeneration now responds to changes in the Redis sidecar configuration.

* **Tests**
  * Added tests for sidecar cache config, legacy cache fallback, and Unix-socket health checks.

### DIFF
--- a/aap_gateway_api/defaults.py
+++ b/aap_gateway_api/defaults.py
@@ -20,6 +20,8 @@ AAP_DEPLOYMENT_TYPE = "self-managed"
 
 ALLOWED_HOSTS = ["*"]
 
+GATEWAY_ENABLE_LEGACY_CACHE = False
+
 # This setting should not be overrideable
 AOC_UNCHANGEABLE_PREFERENCES = ['gateway_token_name']
 
@@ -80,6 +82,17 @@ CACHES = {
     "fallback": {
         "BACKEND": "django.core.cache.backends.filebased.FileBasedCache",
         "LOCATION": "/var/tmp/django_cache",
+    },
+    "sidecar": {
+        "BACKEND": "django_redis.cache.RedisCache",
+        "LOCATION": "unix:///var/run/redis/redis.sock?db=0",
+        "KEY_PREFIX": "gateway",
+        "OPTIONS": {
+            "CLIENT_CLASS": "django_redis.client.DefaultClient",
+            "CONNECTION_POOL_KWARGS": {
+                "protocol": 2,
+            },
+        },
     },
 }
 
--- a/aap_gateway_api/defaults.py
+++ b/aap_gateway_api/defaults.py
@@ -20,8 +20,6 @@ AAP_DEPLOYMENT_TYPE = "self-managed"
 
 ALLOWED_HOSTS = ["*"]
 
-GATEWAY_ENABLE_LEGACY_CACHE = False
-
 # This setting should not be overrideable
 AOC_UNCHANGEABLE_PREFERENCES = ['gateway_token_name']
 
@@ -82,17 +80,6 @@ CACHES = {
     "fallback": {
         "BACKEND": "django.core.cache.backends.filebased.FileBasedCache",
         "LOCATION": "/var/tmp/django_cache",
-    },
-    "sidecar": {
-        "BACKEND": "django_redis.cache.RedisCache",
-        "LOCATION": "unix:///var/run/redis/redis.sock?db=0",
-        "KEY_PREFIX": "gateway",
-        "OPTIONS": {
-            "CLIENT_CLASS": "django_redis.client.DefaultClient",
-            "CONNECTION_POOL_KWARGS": {
-                "protocol": 2,
-            },
-        },
     },
 }
 
--- a/aap_gateway_api/settings_utils.py
+++ b/aap_gateway_api/settings_utils.py
@@ -37,7 +37,6 @@ _CUSTOM_ENVVAR_MAPPINGS = (
     ("REDIS_CERT_FILE", "CACHES__primary__OPTIONS__CLIENT_CLASS_KWARGS__ssl_certfile"),
     ("REDIS_CA_CERT_FILE", "CACHES__primary__OPTIONS__CLIENT_CLASS_KWARGS__ssl_ca_certs"),
     ("FALLBACK_CACHE_FILE", "CACHES__fallback__LOCATION"),
-    ("GATEWAY_ENABLE_LEGACY_CACHE", "GATEWAY_ENABLE_LEGACY_CACHE", to_python_boolean),
     ("CSRF_TRUSTED_ORIGINS", "CSRF_TRUSTED_ORIGINS"),
     ("LOGOUT_ALLOWED_HOSTS", "LOGOUT_ALLOWED_HOSTS", lambda v: v.split(",")),
     ("PING_PAGE_CHECK_TIMEOUT", "PING_PAGE_CHECK_TIMEOUT"),
@@ -68,16 +67,6 @@ def load_custom_envvars(settings):
         data["GRPC_SERVER_MAX_RECEIVE_MESSAGE_LENGTH"] = settings.GRPC_SERVER_MAX_RECEIVE_MESSAGE_LENGTH_MIN_VALUE
 
     settings.update(data, loader_identifier="settings:load_custom_envvars", merge=True)
-
-    if not settings.get("GATEWAY_ENABLE_LEGACY_CACHE", False):
-        sidecar_config = settings.get("CACHES", {}).get("sidecar")
-        if sidecar_config:
-            logger.info("Sidecar cache override: replacing primary cache with sidecar config")
-            settings.set(
-                "CACHES__primary",
-                sidecar_config,
-                loader_identifier="settings:sidecar_cache_override",
-            )
 
 
 def set_secret_key(settings):
--- a/aap_gateway_api/settings_utils.py
+++ b/aap_gateway_api/settings_utils.py
@@ -37,6 +37,7 @@ _CUSTOM_ENVVAR_MAPPINGS = (
     ("REDIS_CERT_FILE", "CACHES__primary__OPTIONS__CLIENT_CLASS_KWARGS__ssl_certfile"),
     ("REDIS_CA_CERT_FILE", "CACHES__primary__OPTIONS__CLIENT_CLASS_KWARGS__ssl_ca_certs"),
     ("FALLBACK_CACHE_FILE", "CACHES__fallback__LOCATION"),
+    ("GATEWAY_ENABLE_LEGACY_CACHE", "GATEWAY_ENABLE_LEGACY_CACHE", to_python_boolean),
     ("CSRF_TRUSTED_ORIGINS", "CSRF_TRUSTED_ORIGINS"),
     ("LOGOUT_ALLOWED_HOSTS", "LOGOUT_ALLOWED_HOSTS", lambda v: v.split(",")),
     ("PING_PAGE_CHECK_TIMEOUT", "PING_PAGE_CHECK_TIMEOUT"),
@@ -67,6 +68,15 @@ def load_custom_envvars(settings):
         data["GRPC_SERVER_MAX_RECEIVE_MESSAGE_LENGTH"] = settings.GRPC_SERVER_MAX_RECEIVE_MESSAGE_LENGTH_MIN_VALUE
 
     settings.update(data, loader_identifier="settings:load_custom_envvars", merge=True)
+
+    if not settings.get("GATEWAY_ENABLE_LEGACY_CACHE", False):
+        sidecar_config = settings.get("CACHES", {}).get("sidecar")
+        if sidecar_config:
+            settings.set(
+                "CACHES__primary",
+                sidecar_config,
+                loader_identifier="settings:sidecar_cache_override",
+            )
 
 
 def set_secret_key(settings):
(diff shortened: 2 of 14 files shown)


aap_gateway_api/defaults.py
```diff
@@ -20,6 +20,8 @@
 
 ALLOWED_HOSTS = ["*"]
 
+GATEWAY_ENABLE_LEGACY_CACHE = False
+
```
john-westcott-iv: We actually don't want this in here. This will be added by users if needed.

aap_gateway_api/defaults.py
```diff
@@ -76,6 +78,17 @@
         "BACKEND": "django.core.cache.backends.filebased.FileBasedCache",
         "LOCATION": "/var/tmp/django_cache",
     },
+    "sidecar": {
```
john-westcott-iv: This wants to change to "default" and make the current "default" called "legacy".

aap_gateway_api/defaults.py
```diff
@@ -76,6 +78,17 @@
...
+        "BACKEND": "django_redis.cache.RedisCache",
+        "LOCATION": "unix:///var/run/redis/redis.sock?db=0",
```
john-westcott-iv: We probably want to use db=4 here if possible (that is what it will be in the installers even though they are going to override this).

aap_gateway_api/settings_utils.py
```diff
@@ -68,6 +69,16 @@ def load_custom_envvars(settings):
...
+    if not settings.get("GATEWAY_ENABLE_LEGACY_CACHE", False):
+        sidecar_config = settings.get("CACHES", {}).get("sidecar")
+        if sidecar_config:
+            logger.info("Sidecar cache override: replacing primary cache with sidecar config")
+            settings.set(
+                "CACHES__primary",
+                sidecar_config,
+                loader_identifier="settings:sidecar_cache_override",
+            )
+
```
john-westcott-iv: Cringe, creative but don't do this here, the drivers are going to implement something to pick between the two.

tools/ansible/roles/sources/templates/docker-compose-stage.yml.j2
```diff
@@ -36,6 +37,7 @@ services:
       - REDIS_HOSTS={{ ",".join(redis_hosts) }}
       - REDIS_TLS={{ redis_uses_tls }}
       - CACHE_KEY_PREFIX={{ cache_key_prefix }}
+      - GATEWAY_ENABLE_LEGACY_CACHE=${GATEWAY_ENABLE_LEGACY_CACHE:-False}
```
john-westcott-iv: We might want to make this a django variable instead of an env var so that it could be set in the startup-config.yml file.

tools/ansible/roles/sources/templates/docker-compose-stage.yml.j2
```diff
@@ -74,12 +76,36 @@ services:
...
+      - redis_sidecar_sock{{ container_id }}:/var/run/redis:z
     ulimits:
       nofile:
         soft: 4096
         hard: 4096
   {% endfor %}
 
+{# Sidecar Redis — one per gateway, Unix socket only, no TLS/ACL #}
+{% for i in range(gateway_node_count|int) %}
+{% set container_id = loop.index %}
```
john-westcott-iv: Instead of duplicating this for loop could we just move this content into the for loop for the main gateway servers since this will always be 1:1?

tools/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -35,6 +36,7 @@ services:
       - REDIS_HOSTS={{ ",".join(redis_hosts) }}
       - REDIS_TLS={{ redis_uses_tls }}
       - CACHE_KEY_PREFIX={{ cache_key_prefix }}
+      - GATEWAY_ENABLE_LEGACY_CACHE=${GATEWAY_ENABLE_LEGACY_CACHE:-False}
```
john-westcott-iv: Same comment as above.

tools/ansible/roles/sources/templates/docker-compose.yml.j2
```diff
@@ -83,12 +85,36 @@ services:
...
+      - redis_sidecar_sock{{ container_id }}:/var/run/redis:z
     ulimits:
       nofile:
         soft: 4096
         hard: 4096
   {% endfor %}
 
+{# Sidecar Redis — one per gateway, Unix socket only, no TLS/ACL #}
+{% for i in range(gateway_node_count|int) %}
+{% set container_id = loop.index %}
```
john-westcott-iv: Same comment as above.

tools/ansible/roles/sources/templates/redis-sidecar.conf.j2
```diff
@@ -0,0 +1,12 @@
...
+unixsocketperm 777
+protected-mode no
+maxmemory 128mb
+maxmemory-policy allkeys-lru
+save ""
+appendonly no
+loglevel notice
+tcp-keepalive 0
+timeout 0
+databases 1
```
john-westcott-iv: This might need to change if you switch to DB 4.

john-westcott-iv (dismissed):
Several design alignment issues — see inline comments. The sidecar approach is right, but the integration points need to align with the driver-level cache selection that's coming.

aap_gateway_api/settings_utils.py
jeffh-oss: Removed

tools/ansible/roles/sources/templates/docker-compose-stage.yml.j2
jeffh-oss: Makes sense. Moved.

tools/ansible/roles/sources/templates/redis-sidecar.conf.j2
jeffh-oss: Changed to 5

aap_gateway_api/defaults.py
jeffh-oss: Using 4 now

tools/ansible/roles/sources/templates/docker-compose-stage.yml.j2
jeffh-oss: Removed as it will be handled in a future PR for the drivers work

aap_gateway_api/defaults.py
jeffh-oss: default is now the sidecar and the old approach is called legacy

jay-steurer:
**Heads up: this change breaks the `/status` endpoint Redis health check**

The `check_redis()` function in `status.py` selects which cache to health-check based on whether `"default"` uses `DABCacheWithFallback`. After this PR, `"default"` is the sidecar, so it tries to health-check the Unix socket instead of the centralized Redis, causing the status endpoint to report Redis as failed in non-dev environments.


_Assisted-by Claude AI_

aap_gateway_api/tests/views/api/v1/status/test_status.py
```diff
@@ -491,5 +491,88 @@ def test_null_ping_url(get, admin_api_client):
...
+        },
+        "primary": {
+            "BACKEND": "django_redis.cache.RedisCache",
+            "LOCATION": "rediss://gateway1:redis_password@redis:6380",
+            "OPTIONS": {
+                "CLIENT_CLASS": "ansible_base.lib.redis.RedisClient",
+                "CLIENT_CLASS_KWARGS": {"mode": "standalone"},
+            },
+        },
+    }
```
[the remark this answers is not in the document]
jeffh-oss: Claude says

> It's technically valid (avoids drift if the constant ever changes), but it's a minor style point. The existing tests throughout the file also hardcode "primary".

So I'm not going to do this at this time unless the humans feel we should switch to using a constant for this.

aap_gateway_api/views/api/v1/status.py
```diff
@@ -34,12 +35,32 @@ def check_redis(timeout: int = 4) -> Dict:
...
+    try:
+        parsed = redis_lib.connection.parse_url(url)
+        r = redis_lib.Redis(
+            unix_socket_path=parsed.get('path', '/var/run/redis/redis.sock'),
+            db=parsed.get('db', 0),
+            socket_timeout=timeout,
+            protocol=2,
+        )
+        r.ping()
+        return {'mode': 'standalone', 'status': STATUS_GOOD}
```
bhavenst: Nit: `r` is never closed — each `/status` call creates a new `redis.Redis()` with its own connection pool, and the Unix socket fd is only reclaimed when GC collects it.

The upstream `get_redis_status()` has the same pattern, so this is consistent — but since this is new code, adding a `finally: r.close()` is trivial and prevents fd accumulation under repeated health polling:

```python
def _check_redis_unix_socket(url: str, timeout: int) -> Dict:
    try:
        parsed = redis_lib.connection.parse_url(url)
        r = redis_lib.Redis(
            unix_socket_path=parsed.get('path', '/var/run/redis/redis.sock'),
            db=parsed.get('db', 0),
            socket_timeout=timeout,
            protocol=2,
        )
        try:
            r.ping()
            return {'mode': 'standalone', 'status': STATUS_GOOD}
        finally:
            r.close()
    except Exception as e:
        logger.exception("Failed checking sidecar Redis health")
        return {'mode': 'standalone', 'status': STATUS_FAILED, 'exception': str(e)}
```

jeffh-oss: Agree and Claude apologizes for the poor code

> Fixed. The r.close() now runs in a finally block so the Redis connection is cleaned up whether ping() succeeds or raises. Fair point — I should have written it that way from the start.

bhavenst (approved):
Re-review: feedback addressed — Redis connection properly closed. LGTM.

tznamena (approved):
LGTM

PR #107, opened 2026-06-05
[AAP-76814] Short-circuit migrate_service_data when data is already synchronized

huffmanca:
## Summary
- Add early-exit check in `handle()` using `MigrateServiceDataHasRan` flag to skip the entire command when migration has previously completed (also addresses AAP-77663)
- Add per-service unmigrated resource count check in `_migrate_single_service()` to skip services where all resources have already been migrated, handling the crash-before-flag-set edge case
- Add tests for both short-circuit and proceed-normally paths

## Context

The `migrate_service_data` command runs on every operator reconcile loop with no guard for whether migration has already completed. At scale (100k+ resources), this causes hours of unnecessary API calls and DB load on every reconcile. Related to AAP-77588 (Kyndryl escalation).

Two defense-in-depth layers:
1. **`MigrateServiceDataHasRan` check** — instant bail if the flag is set (covers the normal case)
2. **Per-service unmigrated count check** — queries each resource type filtered by upstream `service_id`; if all return count=0, the service is already synced (covers the crash-before-flag-set case)

## Test plan
- [ ] Existing tests pass (`test_migrate_resources.py`)
- [ ] New test: `test_migration_skips_when_already_synced` — verifies short-circuit fires when count=0
- [ ] New test: `test_migration_proceeds_when_not_synced` — verifies migration proceeds when count>0
- [ ] Manual: verified on aap-dev cluster — second run drops from 4m11s to 5s

## Summary by CodeRabbit

* **New Features**
  * Migration now detects and skips services already synchronized, avoiding redundant resource migration steps
  * Command exits early if the overall migration flag indicates completion
  * Role-assignment migration for users and teams continues to run regardless of resource sync state

* **Tests**
  * Added tests and an autouse fixture to verify skip behavior when services are synchronized and normal execution when not

### DIFF
--- a/aap_gateway_api/management/commands/migrate_service_data.py
+++ b/aap_gateway_api/management/commands/migrate_service_data.py
@@ -134,6 +134,10 @@ class Command(BaseCommand):
         """
         self._warn_ignored_flags(options)
 
+        if MigrateServiceDataHasRan.has_migration_completed():
+            self.stdout.write("Migration has already completed. Skipping.")
+            return
+
         # Force merge options to True as per requirements
         merge_teams = True
         merge_organizations = True
@@ -303,6 +307,19 @@ class Command(BaseCommand):
                 continue
         return failed_services
 
+    def _is_service_already_synced(self) -> bool:
+        """Check if all resource types for the current service have already been migrated."""
+        for resource_type_name in self.resource_types_to_migrate:
+            filters = {
+                "service_id": self.upstream_service_id,
+                "is_partially_migrated": "false",
+                "content_type__resource_type__name": resource_type_name,
+            }
+            data = self.client.list_resources(filters=filters).json()
+            if data["count"] > 0:
+                return False
+        return True
+
     def _migrate_single_service(
         self,
         service_api: ServiceAPIRoute,
@@ -349,6 +366,10 @@ class Command(BaseCommand):
 
         service_api.service_cluster.service_id = self.upstream_service_id
         service_api.service_cluster.save()
+
+        if self._is_service_already_synced():
+            self.stdout.write(f"Service {service_slug} is already synchronized — no unmigrated resources found. Skipping migration.")
+            return True, None
 
         self.stdout.write(
             f"Migrating {', '.join(self.resource_types_to_migrate.keys())} from {upstream_service_type}, id: {self.upstream_service_id} into Gateway"
--- a/aap_gateway_api/management/commands/migrate_service_data.py
+++ b/aap_gateway_api/management/commands/migrate_service_data.py
@@ -367,19 +367,18 @@ class Command(BaseCommand):
         service_api.service_cluster.service_id = self.upstream_service_id
         service_api.service_cluster.save()
 
-        if self._is_service_already_synced():
-            self.stdout.write(f"Service {service_slug} is already synchronized — no unmigrated resources found. Skipping migration.")
-            return True, None
-
-        self.stdout.write(
-            f"Migrating {', '.join(self.resource_types_to_migrate.keys())} from {upstream_service_type}, id: {self.upstream_service_id} into Gateway"
-        )
-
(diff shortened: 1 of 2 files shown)


aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -350,6 +367,10 @@ def _migrate_single_service(
         service_api.service_cluster.service_id = self.upstream_service_id
         service_api.service_cluster.save()
 
+        if self._is_service_already_synced():
+            self.stdout.write(f"Service {service_slug} is already synchronized — no unmigrated resources found. Skipping migration.")
+            return True, None
```
bhavenst: This early return also skips `migrate_role_assignments()` (lines 384-385) and `delete_legacy_authenticators()` (line 379). In the crash-after-resource-migration scenario this PR is designed to handle, role assignments would be permanently lost: all services report success, `mark_migration_completed()` gets called, and future runs hit the global early-exit.

Suggest moving the guard to wrap only the resource loop:
```python
if not self._is_service_already_synced():
    self.delete_legacy_authenticators()
    for r_type in self.resource_types_to_migrate.keys():
        self.migrate_resource(r_type)
else:
    self.stdout.write(f"Service {service_slug} is already synchronized -- skipping resource migration.")

self.migrate_role_assignments(AssignmentActorType.USER, service_slug, service_type_name)
self.migrate_role_assignments(AssignmentActorType.TEAM, service_slug, service_type_name)
```

bhavenst (changes requested):
Short-circuit for the global `MigrateServiceDataHasRan` flag looks correct. The per-service `_is_service_already_synced()` guard has a correctness issue — see inline.

bhavenst (approved):
Feedback addressed — the per-service sync check now wraps only the resource migration loop, so `delete_legacy_authenticators()` and `migrate_role_assignments()` run unconditionally. The new test explicitly validates that role assignments still execute when resources are skipped. Test isolation fixture and mock fixes look good. LGTM.

aap_gateway_api/management/commands/migrate_service_data.py
huffmanca: Valid, but unfortunate. The role assignments is where we spend the vast majority of time, and I agree that we shouldn't risk the scenario where we lose any unsynced role assignments.

So now we have the scenarios of:
- successful sync, we skip in a couple of seconds (due to the `has_migration_completed` check before this)
- the crash scenario, we'll still spend a fair bit of time processing through these assignments once more.

The original Jira issue suggested doing a `count()` and skipping if the counts matched. I thought checking the sync status was better for the entities above, but the role assignments don't have the `service_id` or `is_migrated` flags, so we can't take that approach here.

Do you think we should add in the `count()` check and skip if the counts match between components, even though that has its own potential edge cases?

bhavenst: A count-based skip for role assignments is risky — gateway-side counts include assignments from all services plus natively-created ones, and upstream queries apply role definition exclusion filters that can't be replicated on the gateway side. A false positive would leave users without permissions.

The current design handles this well: `MigrateServiceDataHasRan` makes the normal case instant, and `give_permission`/`give_global_permission` are idempotent (`get_or_create` under the hood), so re-running in the crash recovery case is safe — just slower. Given how rare the crash scenario is, the re-run cost seems acceptable.

If this becomes a bottleneck in practice, the right long-term fix would be adding a `source_service_id` field to `RoleUserAssignment`/`RoleTeamAssignment` in DAB and using the same pattern as `_is_service_already_synced`. But that's a schema change and probably not worth it unless the perf cost is measurable.

huffmanca: Sounds good. We'll leave this one as-is, though, and work on improving the migrate_service_data sync through the other issues.

PR #111, opened 2026-06-09
[AAP-78372] Fix ping endpoint hang during DB outage

Dostonbek1:
## Description

* **What:** Introduce a dedicated `healthcheck` DATABASES alias with `connect_timeout=3` and call `close_old_connections()` before the health-check query in `PingView._check_db()`.
* **Why:** During a database restart, `PingView._check_db()` calls `connection.cursor()` which attempts a new TCP connection. If the DB host is network-unreachable, the TCP `connect()` blocks for ~130s (default Linux SYN retransmission timeout), far exceeding the K8s liveness probe's 5s `timeoutSeconds`. The `try/except` never fires because the code is stuck inside the connection attempt. After 5 consecutive probe timeouts (75s), Kubernetes kills the pod, triggering a crashloop with exponential backoff that extends the outage to 10-15 minutes.
* **How:** Two changes:
  1. A `healthcheck` DATABASES alias is created in `load_healthcheck_settings()` (called from `settings.py` after all installer overrides are applied). It deep-copies the `default` DB config via `to_dict()` and overlays `connect_timeout=3`, `CONN_MAX_AGE=0`, and `CONN_HEALTH_CHECKS=True`.
  2. `PingView._check_db()` calls `close_old_connections()` then runs `SELECT 1` on the `healthcheck` alias. This matches the existing pattern in `proxy/control_plane.py:_ensure_db_availability` (added in AAP-42378). Single attempt caps worst-case at ~3s, within the 5s probe budget.

## Type of Change

- Bug fix (non-breaking change which fixes an issue)
- Test update

## Self-Review Checklist

### Prerequisites

* Operator-based deployment with an external PostgreSQL database

### Steps to Test

1. Deploy and verify gateway is healthy via `/api/gateway/v1/ping/`
2. Trigger a DB restart (e.g. `pg_ctl restart` or cloud provider restart)
3. Monitor gateway pod behavior during and after the outage

### Expected Results

* During DB outage: ping returns HTTP 200 with `{"status": "degraded", "db_connected": false}` within 3 seconds (instead of hanging)
* Liveness probe continues passing (pod is NOT killed)
* Readiness probe fails (pod removed from endpoints — correct behavior)
* After DB recovers: gateway reconnects and becomes Ready within 1-2 minutes (no manual pod restart needed)

### Unit Tests

* `test_ping_db_check_flushes_stale_connections` verifies ordering: `close_old_connections()` is called before `connections["healthcheck"].cursor()`
* All existing ping tests pass with the autouse mock fixture (matching the `test_proxy.py` pattern)

### Required Actions

* Requires downstream repository changes
  * Longer term: The operator should separate liveness from readiness probes. Liveness should not check DB at all — it should only verify the WSGI process can respond.

### Related Issues

* Fixes AAP-78372
* Related: AAP-42378 (previous partial fix — gRPC path only)

---

### DIFF
--- a/aap_gateway_api/settings_utils.py
+++ b/aap_gateway_api/settings_utils.py
@@ -102,6 +102,25 @@ def load_grpc_settings(settings: Dynaconf) -> None:
     load_python_file_with_injected_context(settings_file_path, settings=settings)
 
 
+def load_healthcheck_settings(settings: Dynaconf) -> None:
+    """Create a 'healthcheck' DATABASES alias derived from 'default'.
+
+    Deep-copies the full default DB config via ``to_dict()`` so every key
+    (including any installer-added ones) is preserved, then overlays an
+    aggressive connect_timeout so PingView._check_db() fails fast instead
+    of blocking for ~130 s on unreachable hosts.
+    """
+    default_db = settings.DATABASES.get("default")
+    if not default_db:
+        return
+    healthcheck_db = default_db.to_dict()
+    healthcheck_db.setdefault("OPTIONS", {})["connect_timeout"] = 3
+    healthcheck_db["CONN_MAX_AGE"] = 0
+    healthcheck_db["CONN_HEALTH_CHECKS"] = True
+    healthcheck_db["TEST"] = {"MIRROR": "default"}
+    settings.set("DATABASES__healthcheck", healthcheck_db)
+
+
 def load_oidc_provider_settings(settings: Dynaconf) -> None:
     """Load OIDC provider settings.
 
--- a/aap_gateway_api/settings.py
+++ b/aap_gateway_api/settings.py
@@ -2,7 +2,14 @@ import os
 
 from ansible_base.lib.dynamic_config import export, factory, load_envvars, load_python_file_with_injected_context, load_standard_settings_files
 
-from .settings_utils import _GATEWAY_ETC_DIRECTORY, load_custom_envvars, load_grpc_settings, load_oidc_provider_settings, set_secret_key
+from .settings_utils import (
+    _GATEWAY_ETC_DIRECTORY,
+    load_custom_envvars,
+    load_grpc_settings,
+    load_healthcheck_settings,
+    load_oidc_provider_settings,
+    set_secret_key,
+)
 
 DYNACONF = factory(
     __name__,
@@ -23,6 +30,7 @@ load_standard_settings_files(DYNACONF)  # /etc/ansible-automation-platform/*.yam
 load_custom_envvars(DYNACONF)  # load custom unprefixed envvars
 load_envvars(DYNACONF)  # load envvars prefixed with GATEWAY_
 set_secret_key(DYNACONF)  # set secret key based on secret key file
+load_healthcheck_settings(DYNACONF)  # create healthcheck DB alias from default
 
 load_oidc_provider_settings(DYNACONF)  # conditionally enable Gateway as OIDC Provider
 
--- a/aap_gateway_api/tests/views/api/v1/ping/test_ping.py
+++ b/aap_gateway_api/tests/views/api/v1/ping/test_ping.py
@@ -9,6 +9,25 @@ from aap_gateway_api.models import HTTPPort, ServiceAPIRoute
 from aap_gateway_api.version import get_aap_version
 
 
+@pytest.fixture(autouse=True)
+def _mock_ping_db_access():
+    """Mock close_old_connections and redirect healthcheck alias to the default connection.
+
+    close_old_connections would close the test DB connection, so we mock it away
+    (same pattern as test_proxy.py).  The healthcheck DATABASES alias is a separate
+    connection that @pytest.mark.django_db does not grant access to, so we redirect
+    it to the default connection which the test framework manages.
+    """
+    from django.db import connection
+
+    with (
+        mock.patch("aap_gateway_api.views.api.v1.ping.close_old_connections"),
+        mock.patch("aap_gateway_api.views.api.v1.ping.connections") as mock_conns,
+    ):
+        mock_conns.__getitem__.return_value = connection
+        yield
+
+
 @pytest.mark.django_db
 @mock.patch("aap_gateway_api.views.api.v1.ping.PingView._check_dispatcherd", return_value=True)
 @mock.patch("aap_gateway_api.views.api.v1.ping.requests.request")
@@ -80,6 +99,27 @@ def test_ping_proxy_non_200(request, mock_dispatcherd, unauthenticated_api_clien
     assert response.status_code == 200
     assert response.data['status'] == STATUS_DEGRADED
     assert response.data['proxy_status_code'] == 500
+
+
+@pytest.mark.django_db
+@mock.patch("aap_gateway_api.views.api.v1.ping.PingView._check_dispatcherd", return_value=True)
+@mock.patch("aap_gateway_api.views.api.v1.ping.requests.request")
+@mock.patch("aap_gateway_api.views.api.v1.ping.connections")
+@mock.patch("aap_gateway_api.views.api.v1.ping.close_old_connections")
+def test_ping_db_check_flushes_stale_connections(mock_close, mock_conns, request, mock_dispatcherd, unauthenticated_api_client):
+    """close_old_connections must run before the healthcheck cursor (AAP-78372)."""
+    request.return_value = mock.Mock(status_code=200, json=lambda: {"test": "test"})
+    call_order = []
+    mock_close.side_effect = lambda: call_order.append("close")
+    cursor_cm = mock.MagicMock()
+    cursor_cm.__enter__ = mock.Mock(side_effect=lambda: (call_order.append("cursor") or mock.Mock()))
+    cursor_cm.__exit__ = mock.Mock(return_value=False)
+    mock_conns.__getitem__.return_value.cursor.return_value = cursor_cm
+
+    url = get_relative_url("ping-view")
+    response = unauthenticated_api_client.get(url)
+    assert response.status_code == 200
+    assert call_order == ["close", "cursor"]
 
 
 @pytest.mark.django_db
--- a/aap_gateway_api/views/api/v1/ping.py
+++ b/aap_gateway_api/views/api/v1/ping.py
@@ -4,7 +4,7 @@ from datetime import datetime
 import requests
 from ansible_base.lib.constants import STATUS_DEGRADED, STATUS_GOOD
 from django.conf import settings
-from django.db import connection
+from django.db import close_old_connections, connections
 from rest_framework.response import Response
 
 from aap_gateway_api.models import HTTPPort
@@ -20,7 +20,8 @@ class PingView(AnsibleBaseView):
     serializer_class = PingSerializer
 
     def _check_db(self):
-        with connection.cursor() as cursor:
+        close_old_connections()
+        with connections["healthcheck"].cursor() as cursor:
             cursor.execute("SELECT 1")
 
     def _check_dispatcherd(self):


jeffh-oss:
## Code Review: [AAP-78372] Fix ping endpoint hang during DB outage

**PR:** https://github.com/ansible/jewel/pull/111
**Author:** Doston Toirov (@Dostonbek1)
**Branch:** fix/ping-db-connect-timeout-AAP-78372 → devel
**Commit:** ad174736
**Date:** 2026-06-09
**Review iteration:** 1

### Verdict: READY_FOR_HUMAN_REVIEW

### Scores
| Lens | Score | Findings |
|---|---|---|
| Functionality | 10/10 | 0 Critical, 0 Major, 0 Minor, 0 Nit |
| Security | 10/10 | 0 Critical, 0 Major, 0 Minor, 0 Nit |
| Quality | 9.5/10 | 0 Critical, 0 Major, 1 Minor, 0 Nit |
| **Overall** | **9.8/10** | |

### Findings

#### Critical
None

#### Major
None

#### Minor

1. **[Quality]** **[aap_gateway_api/tests/views/api/v1/ping/test_ping.py:88-96]** Test does not verify call ordering — only verifies `close_old_connections()` was called, not that it was called *before* `connection.cursor()`
   - **Evidence**: `mock_close.assert_called_once()` — the test docstring says "close_old_connections must run before the DB health check" but the assertion only checks presence, not ordering. A future refactor moving `close_old_connections()` after `connection.cursor()` would pass this test but break the fix.
   - **Confidence**: HIGH
   - **Fix**: Use side-effect tracking to verify ordering:
     ```python
     @mock.patch("aap_gateway_api.views.api.v1.ping.connection")
     @mock.patch("aap_gateway_api.views.api.v1.ping.close_old_connections")
     def test_ping_db_check_flushes_stale_connections(mock_close, mock_conn, request, mock_dispatcherd, unauthenticated_api_client):
         """close_old_connections must run before the DB health check (AAP-78372)."""
         call_order = []
         mock_close.side_effect = lambda: call_order.append("close")
         mock_conn.cursor.return_value.__enter__ = mock.Mock(
             side_effect=lambda: call_order.append("cursor") or mock.Mock()
         )
         mock_conn.cursor.return_value.__exit__ = mock.Mock(return_value=False)
         request.return_value = mock.Mock(status_code=200, json=lambda: {"test": "test"})
         url = get_relative_url("ping-view")
         response = unauthenticated_api_client.get(url)
         assert response.status_code == 200
         assert call_order == ["close", "cursor"]
     ```
   - **Points**: 0.5

#### Nit
None

### Needs Human Judgment

- **Jira context unavailable:** Could not fetch AAP-78372 or AAP-42378 details from Jira (no Jira credentials configured). The PR description provides thorough context, but human reviewers should verify alignment with the Jira acceptance criteria directly.

### Verification Results
| connect_timeout consistency | `grep -rn "connect_timeout" aap_gateway_api/` | `defaults.py:75: 'socket_connect_timeout': 1` (Redis, pre-existing), `defaults.py:115: "connect_timeout": 3` (DB, new) | PASS — no conflicting values |
| close_old_connections pattern | `grep -rn "close_old_connections" aap_gateway_api/` | Used identically in `proxy/control_plane.py:17,176` (import + call) | PASS — matches existing codebase pattern |
| _check_db callers | `grep -rn "_check_db" aap_gateway_api/` | Only called at `views/api/v1/ping.py:46`, mocked in `test_ping.py:35` | PASS — single call site, no missed callers |
| DATABASES definition | `grep -rn "DATABASES" aap_gateway_api/defaults.py` | Single definition at line 104 | PASS — no duplicate definitions |

### Observations

1. **Follows established codebase patterns.** The `close_old_connections()` call mirrors the exact same pattern in `aap_gateway_api/proxy/control_plane.py:176` (`_ensure_db_availability`). The PR description explicitly references this (AAP-42378).

2. **`connect_timeout` aligns with existing timeout strategy.** Redis cache uses `socket_connect_timeout: 1`, `PING_PAGE_CHECK_TIMEOUT` is already `3`, and the K8s liveness probe has a 5s timeout. The 3-second DB connect timeout fits within this budget.

3. **Interaction with `CONN_HEALTH_CHECKS: True` is complementary, not redundant.** `CONN_HEALTH_CHECKS` verifies persistent connections before reuse but doesn't close them. `close_old_connections()` actively closes stale connections. Both are needed because a half-open TCP socket will fail the health check, but the re-establishment `connect()` could still block for 130s without `connect_timeout`.

4. **No concurrency concerns.** Django's `connection` is thread-local. `close_old_connections()` only affects the current thread's connections.

5. **Error handling is adequate.** If `close_old_connections()` raises, the exception propagates through the `except Exception` block at line 69, setting `STATUS_DEGRADED`. No silent failure path.

6. **No new information disclosure.** The diff does not change what is returned in the API response. Pre-existing exposure of `db_exception` type names is unchanged.

7. **No injection risk.** `connect_timeout` is a hardcoded integer literal, not derived from any external input. `close_old_connections()` is a parameterless Django standard library call.

8. **No dependency changes.** `close_old_connections` is from `django.db`, already a project dependency.

9. **Test follows existing patterns.** The mock patching style and fixture usage match the existing test file conventions.

### Incidental Findings (out of scope — not scored)

1. **Pre-existing: `datetime.now()` without timezone (ping.py:38).** The `pong` field uses `datetime.now()` which returns a naive datetime. In a timezone-aware Django project (`USE_TZ = True` in defaults.py) this could produce warnings. Not introduced by this diff.

### Limitations
This review was performed by an AI agent. It does not understand business context, domain intent, organizational constraints, or deployment environment specifics. LOW-confidence findings and items in "Needs Human Judgment" require human verification. This review is a first pass, not a final approval.

### Validation Summary
| Functionality | 0 | 0 | 0 | 0 |
| Security | 0 | 0 | 0 | 0 |
| Quality | 1 | 0 | 0 | 1 (Minor) |
| **Total** | **1** | **0** | **0** | **1** |

### Path to 10/10
Each item is an agent-ready fix prompt — paste into your coding agent to apply:

1. **[Quality] [aap_gateway_api/tests/views/api/v1/ping/test_ping.py:88-96]** (+0.5 points)
   > In `test_ping.py` at lines 88-96, `test_ping_db_check_flushes_stale_connections` only asserts `mock_close.assert_called_once()` which verifies presence but not ordering.
   > Add side-effect tracking to verify `close_old_connections()` is called before `connection.cursor()`, or alternatively add an explanatory comment in `_check_db()` documenting why ordering matters for future maintainers.

### Final Score
**Overall: 9.8/10 — READY_FOR_HUMAN_REVIEW**

---


[AAP-78372]: https://redhat.atlassian.net/browse/AAP-78372?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

john-westcott-iv (changes requested):
## Code Review: [AAP-78372] Fix ping endpoint hang during DB outage

The direction here is right — `connect_timeout` and flushing stale connections are the correct tools. A few refinements before this is ready to merge.

### Question: Understanding the timeline

We've reviewed the material and agree that the code has been vulnerable since 09/2025, but we'd like to get a better understanding of why we haven't seen this error until now (9 months later). It could just be "dumb luck" that it hasn't happened, but we should work with other teams to make sure nothing else has changed in the topology being deployed that this fix could potentially mask before we have an understanding of the root of the issue.

### Major: Global `connect_timeout` will be silently overwritten by all installers

We checked all three installer repos — containerized, RPM, and OCP operator — and all of them overwrite the entire `DATABASES` dict in their generated `settings.py`. The `connect_timeout: 3` in `defaults.py` will be lost in every deployed environment:

| Installer | File | Has connect_timeout? |
|-----------|------|---------------------|
| containerized | `roles/automationgateway/templates/settings.py.j2` | No |
| RPM | `roles/automationgateway/templates/settings.py.j2` | No |
| OCP operator (dev/bootstrap) | `configmap_settings.yaml.j2` | Yes, but `10` not `3` |

**Suggestion:** Consider a dedicated `healthcheck` DATABASES alias with `connect_timeout: 3` and `CONN_MAX_AGE: 0` instead of modifying the global default. This follows the existing pattern in `grpc_defaults.py` (which already differentiates DB OPTIONS per use case), isolates the aggressive timeout to the ping endpoint, and avoids the installer override problem entirely.

### Major: Retry pattern can exceed K8s probe timeout

The try/except/retry in `_check_db()` can take up to 6s in the worst case (two 3s TCP connection attempts), exceeding the 5s K8s liveness probe timeout — the exact problem this PR is fixing.

**Suggestion:** Revert to the simpler "call before" approach from the earlier revision: `close_old_connections()` then one check attempt. This caps worst-case at ~3s and matches the pattern in `proxy/control_plane.py:176`.

### Nit: PR description doesn't match current code

The description says "call `close_old_connections()` before the health check query" but the code does a try/except/retry. Worth updating whichever direction you go.

Dostonbek1:
### Local Validation

Patched a running gateway instance with this branch and verified:

| Test | Result |
|---|---|
| Healthcheck DB alias created | `DATABASES` contains `healthcheck` with correct settings |
| `connect_timeout=3` applied | Confirmed in `OPTIONS` |
| `CONN_MAX_AGE=0` | Connections not cached between requests |
| `_check_db()` with DB up | 0.01s |
| `_check_db()` with DB stopped | 0.04s (fast fail) |
| Stale connection recovery | `close_old_connections()` detects broken conn, reconnects in 0.01s |
| Unreachable host (blackhole IP) | Old: hangs 130s+ / New: fails in exactly 3s |

The critical scenario — DB host drops off the network — now fails within the liveness probe budget instead of hanging on TCP SYN retransmissions.

tznamena:
## Review Observations

### Existing review feedback status

All three items from the previous review by @john-westcott-iv appear to have been addressed in the current revision:

1. **Global `connect_timeout` overwritten by installers** — Addressed. The code now creates a dedicated `healthcheck` DATABASES alias dynamically via `load_healthcheck_settings()`.
2. **Retry pattern exceeding K8s probe timeout** — Addressed. The code now uses a single-attempt pattern: `close_old_connections()` then one `SELECT 1`. No retry.
3. **PR description mismatch** — Mostly addressed. The description now accurately describes the approach, though item #3 ("Removed the `connect_timeout` from the `default` DATABASES entry in `defaults.py`") still references a change to `defaults.py` that is not present in the current diff — likely stale from a previous revision.

### Design tradeoff: healthcheck alias tests a different code path than the application

The `healthcheck` alias copies all connection parameters from `default` (same host, same DB, same credentials), but uses `CONN_MAX_AGE=0`. This means it opens a **fresh TCP connection** on every probe, while the application's `default` alias reuses persistent connections (`CONN_MAX_AGE > 0`). These are fundamentally different code paths, and they can diverge:

- **Healthcheck succeeds, app is broken:** After a transparent DB failover (e.g., RDS/Aurora), the healthcheck's fresh connection succeeds while the app's cached connections are stale. The `close_old_connections()` call only cleans up connections in the current thread, not the request-handling threads using the `default` alias.
- **Healthcheck fails, app is fine:** On managed DBs with limited `max_connections`, the application already holds established connections. The healthcheck tries to open a new one each probe and could be rejected with "too many connections" while the app is happily serving. This would cause false-negative degraded status.
- **Connection pooler (PgBouncer) scenarios:** The two aliases are distinct pool entries from PgBouncer's perspective. Under pool exhaustion, one could get a slot while the other doesn't.

The `CONN_MAX_AGE=0` choice also adds connection churn: each probe cycle creates and destroys a connection (e.g., 6 pods × probes every 10s = 6 extra connect/disconnect cycles per 10s on the DB).

This is not necessarily a blocker — the fix correctly solves the real 130s-hang problem. But worth acknowledging that the healthcheck now answers "can I establish a new connection to the DB?" rather than "is the connection the app is using working?" Those are different questions, and in constrained environments (managed DB with connection limits, PgBouncer) the answers can differ.

### Minor: test double-mocking

The `_mock_ping_db_access` autouse fixture mocks `connections` and `close_old_connections` for all tests. Then `test_ping_db_check_flushes_stale_connections` patches them again on top, creating a double-mock stack. It works (the test-specific patches shadow the autouse ones), but it's a bit confusing. Consider either scoping the autouse fixture more narrowly or having the ordering test opt out of it.

tznamena (approved):
Posted a comment about using different DB connection. Not a blocker but we should document the findings somewhere so we and support engineers know that:
- number of connections will go up with this change
- there are chances that application connection and "healthcheck" connections are behaving differently

aap_gateway_api/settings_utils.py
```diff
@@ -102,6 +102,26 @@ def load_grpc_settings(settings: Dynaconf) -> None:
...
+    """
+    default_db = settings.DATABASES.get("default", {})
+    data = {}
+    for key in ("ENGINE", "NAME", "USER", "PASSWORD", "HOST", "PORT"):
+        if key in default_db:
+            data[f"DATABASES__healthcheck__{key}"] = default_db[key]
+    for opt_key, opt_val in default_db.get("OPTIONS", {}).items():
+        data[f"DATABASES__healthcheck__OPTIONS__{opt_key}"] = opt_val
+    settings.update(data, loader_identifier="settings:load_healthcheck_settings", merge=True)
+
```
john-westcott-iv: **Minor: Simplify with `to_dict()` instead of hardcoded key list**

The key-by-key copy is fragile — if someone adds a non-standard top-level key to the default DB config (or a future Django version adds one), it silently won't be copied. Dynaconf's `DynaBox.to_dict()` recursively converts to a fresh plain dict, so you can deep-copy the entire default config in one shot and then overlay the healthcheck-specific settings. This also eliminates the need for the separate `healthcheck_defaults.py` file:

```suggestion
def load_healthcheck_settings(settings: Dynaconf) -> None:
    """Create a 'healthcheck' DATABASES alias derived from 'default'."""
    default_db = settings.DATABASES.get("default")
    if not default_db:
        return
    healthcheck_db = default_db.to_dict()
    healthcheck_db.setdefault("OPTIONS", {})["connect_timeout"] = 3
    healthcheck_db["CONN_MAX_AGE"] = 0
    healthcheck_db["CONN_HEALTH_CHECKS"] = True
    healthcheck_db["TEST"] = {"MIRROR": "default"}
    settings.set("DATABASES__healthcheck", healthcheck_db)
```

john-westcott-iv (commented):
Re-review: the two Major findings from the previous review (installer override of global connect_timeout, retry pattern exceeding probe timeout) are both fully addressed. One new Minor suggestion on simplifying the healthcheck DB alias setup — see inline comment.

john-westcott-iv (approved):

john-westcott-iv:
Re-review complete. All previous findings addressed — the `to_dict()` simplification is clean and eliminates the fragile hardcoded key list and the separate defaults file. No new issues.

One cosmetic nit: PR description item #3 still references a `defaults.py` change that isn't in the diff (stale from an earlier revision) — not blocking.

PR #114, opened 2026-06-12
[AAP-76998] Fix Envoy outlier detection ejecting controller on WebSocket teardowns

aap-platform-services-cicd-bot-sa:
Requires: ansible/ansible.platform#241

## Description

Envoy ejects the controller host after concurrent WebSocket teardowns (`101 UC` / upstream TCP reset). With the default outlier config those UC events count toward `consecutive_5xx` (threshold 5), then all controller traffic gets `503 UH` until uneject.

Split-only (`split_external_local_origin_errors=true`) is not enough. It moves UC into `consecutive_local_origin_failure`, which Envoy still defaults to 5 at 100% enforcement, so the host is still ejected.

This PR:
- Adds `outlier_detection_split_external_local_origin_errors` (default `true`) and emits `split_external_local_origin_errors` in CDS.
- Adds `outlier_detection_consecutive_local_origin_failure` (default `0`) and always emits `consecutive_local_origin_failure` in CDS. When the value is 0 it also emits `enforcing_consecutive_local_origin_failure: 0` so protobuf JSON cannot drop the disable.
- Data migrations set both fields on existing `ServiceCluster` rows.

Active HTTP health checks and `consecutive_5xx` still take out truly unhealthy hosts. This only stops local-origin TCP resets (WebSocket teardowns) from fast-ejecting the cluster.

## Type of Change
- [x] Test update

## Self-Review Checklist

### Prerequisites
A running gateway with this branch (aap-dev or containerized overlay).

### Steps to Test
1. Confirm CDS for the controller cluster includes `splitExternalLocalOriginErrors: true`, `consecutiveLocalOriginFailure: 0`, and `enforcingConsecutiveLocalOriginFailure: 0`.
2. Open 20 connections to `/api/controller/v2/websocket/` and close them gracefully.
3. Probe `/api/controller/v2/ping/` for ~22s (past Envoy's 10s outlier interval).

### Expected Results
Envoy may log `101 UC`. Controller ping/jobs stay HTTP 200. No `503 UH`. Split-only (without consecutive_local_origin_failure=0) still produces `503 UH`.

Verified on aap-dev and on a 2.8 Devel containerized install with this branch overlaid.

### Required Actions
- [x] Requires downstream repository changes
  Collection: ansible/ansible.platform#241. Backport to `aap-gateway` for 2.7.
- [x] Blocked by PR/MR: ansible/ansible.platform#241 (collection completeness; jewel CI checks out this PR)

## JIRA

- Ticket: [AAP-76998](https://redhat.atlassian.net/browse/AAP-76998)

[AAP-76998]: https://redhat.atlassian.net/browse/AAP-76998?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Added controls for distinguishing locally originated failures from external server errors.
  * Added configuration for ejecting hosts after consecutive locally originated failures, disabled by default.
  * Exposed both settings through the service cluster API.
  * Preserved explicit zero values when generating proxy configuration.
  * Documented active health checks as the fallback protection when local-origin ejection is disabled.

* **Bug Fixes**
  * Improved handling of outlier-detection enforcement settings, including zero-value configurations.

* **Tests**
  * Added coverage for defaults, custom thresholds, serialization, and generated proxy configuration.


aap_gateway_api/models/service_cluster.py
```diff
@@ -58,6 +58,24 @@ class ServiceCluster(UniqueNamedCommonModel, AuditableModel):
...
+    )
+
+    outlier_detection_consecutive_local_origin_failure = models.PositiveIntegerField(
+        default=0,
+        help_text=_(
+            "Number of consecutive locally originated failures (connect timeout, TCP reset/UC) before Envoy ejects "
+            "the host. Set to 0 to disable. Takes effect only when split_external_local_origin_errors is true. "
+            "Defaults to 0 so WebSocket teardowns do not eject the cluster."
+        ),
+    )
```
jay-steurer, line 79: Two operator-facing nits on this pair:

1. **Health checks are now the safety net for connect/UC.** These defaults apply to every `ServiceCluster`, not only controller WebSocket. If someone sets `health_checks_enabled=false`, a dead host that fails with connect/`UC` is never ejected (those events no longer count toward `consecutive_5xx`). Worth a sentence in `help_text` here and/or on the health-check field.

2. **`split=false` restores the original bug.** Help text says 0 disables, but that “takes effect only when split is true.” A `PATCH` of split to `false` with consecutive still at `0` looks like disable and then Envoy counts UC toward `consecutive_5xx` again. If `split=false` is not a supported product mode, always emit `true` / reject `false`. If it is supported, say explicitly that turning split off re-enables UC→5xx ejection.

Also: “Set to 0 to disable” is true only if 0 is present on the proto. Unset/omitted → Envoy default **5**. Jewel’s real CDS disable is also `enforcing_consecutive_local_origin_failure: 0` (not a model field). A short mention of that would help operators.

aap_gateway_api/models/route.py
```diff
@@ -154,6 +154,23 @@ def save(self, *args, **kwargs):
...
+            "max_ejection_percent": self.service_cluster.outlier_detection_max_ejection_percent,
+            "split_external_local_origin_errors": self.service_cluster.outlier_detection_split_external_local_origin_errors,
+            "consecutive_local_origin_failure": consecutive_local_origin_failure,
+        }
+        # Envoy's protobuf JSON omits some zero values. enforcing_* defaults to 100, so an
+        # explicit 0 always serializes and disables local-origin ejection even if
+        # consecutive_local_origin_failure=0 is dropped.
+        if consecutive_local_origin_failure == 0:
+            cfg["enforcing_consecutive_local_origin_failure"] = 0
+        return cfg
```
jay-steurer: Emitting `enforcing_*=0` when consecutive is 0 is the right workaround (Envoy unset consecutive → default 5; unset enforcing → 100).

The causal sentence is a bit strong: `MessageToDict` of this packed CDS `Any` **keeps** `consecutiveLocalOriginFailure: 0`. It does omit `splitExternalLocalOriginErrors` when false. Enforcing 0 is still good defense in depth against the **unset → 5** path, not because consecutive 0 necessarily fails to serialize.

Consider moving this helper onto `ServiceCluster` (same idea as `get_effective_health_check_*`) so “0 ⇒ emit enforcing 0” is not only on `Route`.

aap_gateway_api/tests/serializers/test_service_cluster.py
```diff
@@ -32,6 +32,7 @@ def test_serializer_meta_fields_includes_all_cluster_fields(self):
             'outlier_detection_interval_seconds',
             'outlier_detection_base_ejection_time_seconds',
             'outlier_detection_max_ejection_percent',
+            'outlier_detection_consecutive_local_origin_failure',
```
jay-steurer, line 36: This list added `outlier_detection_consecutive_local_origin_failure` but not `outlier_detection_split_external_local_origin_errors` (which is on `Meta.fields`). The test is a subset `assert field in ...`, so dropping split from the serializer would not fail here.

Please add the split field next to this one. The create test below also never sets/asserts split, and consecutive `0` is the model default so it does not prove the input was persisted.

aap_gateway_api/tests/views/api/envoy/test_xds.py
```diff
@@ -114,6 +114,73 @@ def test_xds_outlier_detection_params(admin_api_client, full_service_hierarchy_c
...
+        {"outlier_detection_consecutive_local_origin_failure": threshold},
+    )
+    assert response.status_code == 200
+
+    url = reverse("cds")
+    response = admin_api_client.post(url, data={})
+    assert response.status_code == 200
+
+    outlier_det = response.data['resources'][0]['outlierDetection']
+    assert outlier_det.get('consecutiveLocalOriginFailure', 0) == threshold
```
jay-steurer: Docstring says consecutive is **always present**, including explicit 0. `outlier_det.get('consecutiveLocalOriginFailure', 0) == threshold` still passes if the key is omitted when `threshold == 0`.

The disable that actually serializes is `enforcingConsecutiveLocalOriginFailure == 0` in the default test (no `.get` default) — that part is solid.

Please either:
- `assert outlier_det['consecutiveLocalOriginFailure'] == 0` if CI shows the key is present after proto JSON, or
- change the docstring to: threshold 0 may be omitted; disable is `enforcingConsecutiveLocalOriginFailure == 0`.

jay-steurer (commented):
## Review summary

Not requesting changes. The CDS shape is right: split UC off `consecutive_5xx`, then disable local-origin consecutive ejection with `consecutive_local_origin_failure: 0` plus `enforcing_consecutive_local_origin_failure: 0`. Migration `0026` depends on devel leaf `0025`. Serializer `Meta.fields` includes both new fields. Envoy v1.36.6 treats **unset** consecutive as 5 and **explicit 0** as disable; emitting `enforcing_*=0` is the correct belt-and-suspenders.

Process: completeness is still blocked on ansible/ansible.platform#241 (open; field names match). Konflux ATF was pending at review time.

## Important (nits)

1. **Connect/UC ejection is off for every cluster; health checks are the remaining safety net.** Defaults apply to hub/EDA/gateway as well as controller. If an operator turns `health_checks_enabled` off (supported API), a dead host that fails with connect/`UC` is never ejected. Before this PR those events counted toward `consecutive_5xx`. Spell that out in `help_text`.

2. **`split=false` is a public API value that restores the original bug.** `PATCH` of `outlier_detection_split_external_local_origin_errors=false` with consecutive still at `0` looks like “local-origin eject is disabled.” Envoy then ignores the local-origin knobs and counts UC toward `consecutive_5xx` again. Decide whether `split=false` is a supported product mode; document it (and consider serializer validation) if it is.

3. **Serializer “all fields” test never lists the split boolean.** Subset `assert field in Meta.fields` added `outlier_detection_consecutive_local_origin_failure` but not `outlier_detection_split_external_local_origin_errors`.

4. **CDS tests claim consecutive `0` is always present; `.get(..., 0) == 0` cannot prove that.** The disable that actually serializes is asserted via `enforcingConsecutiveLocalOriginFailure == 0` (no default) — that part is solid. Align the docstring, or use direct key access if CI shows the key is present.

## Suggestions

- `help_text` “Set to 0 to disable” is true only if 0 survives into Envoy. Unset/omitted → Envoy default **5**. Mention that CDS also emits `enforcing_consecutive_local_origin_failure: 0`. The `route.py` comment slightly overstates “JSON omits zeros”; a `MessageToDict` round-trip of this dict **keeps** consecutive 0. Enforcing 0 is defense against the **unset → 5** path.
- `RunPython` is redundant with `AddField(default=True/0)` on existing rows; reverse then `RemoveField` so the reverse update is never observed.
- Consider moving `_get_xds_outlier_detection_config` onto `ServiceCluster` (same idea as `get_effective_health_check_*`).
- If the intent is “no local-origin ejection at all,” also emit `enforcing_local_origin_success_rate: 0` (only matters for clusters with ≥5 hosts).
- Serializer create still sets consecutive `0` (the default) and never sets/asserts split.

## Strengths

Right Envoy levers; `enforcing_*=0` only when consecutive is 0; single `0026` on `0025`; collection option names match; XDS tests are parameterized and follow existing PATCH→CDS style; ruff/tox/migrations/codecov/sonar are green; helper extraction keeps `get_xds_cluster_config` under the complexity limit.

aap_gateway_api/models/service_cluster.py
jeffh-oss: **Note:** This comment was written by Cursor.

Updated help text on both new fields and on `health_checks_enabled`. Split stays a supported opt-out: setting it to false restores counting UC toward `consecutive_5xx`. Consecutive 0 is still the default; the text now says CDS also emits `enforcing_consecutive_local_origin_failure=0` so Envoy does not treat an omitted value as 5, and that health checks are the remaining connect/UC safety net.

aap_gateway_api/models/route.py
jeffh-oss: **Note:** This comment was written by Cursor.

Corrected the comment: consecutive 0 is kept by MessageToDict; `enforcing_*=0` is defense against Envoy treating an omitted consecutive value as 5 (unset enforcing defaults to 100). Left the helper on `Route` for this PR.

aap_gateway_api/tests/serializers/test_service_cluster.py
jeffh-oss: **Note:** This comment was written by Cursor.

Added `outlier_detection_split_external_local_origin_errors` to the expected Meta.fields list. The create test now sets split to false and consecutive to 50 (both non-default) and asserts they persisted.

aap_gateway_api/tests/views/api/envoy/test_xds.py
jeffh-oss: **Note:** This comment was written by Cursor.

Switched to direct key access: `outlier_det['consecutiveLocalOriginFailure']` and `outlier_det['enforcingConsecutiveLocalOriginFailure']`. Local runs of these tests show the consecutive 0 key is present after proto JSON.

jeffh-oss:
**Note:** This comment was written by Cursor.

Addressed the review nits on this PR (replies are also on the inline threads; two of those threads are marked outdated because the lines moved):

- **Help text:** split=false stays a supported opt-out and the text now says it restores UC→5xx ejection. Consecutive 0 also documents `enforcing_consecutive_local_origin_failure=0`. Health checks are called out as the remaining connect/UC safety net.
- **CDS comment:** consecutive 0 is kept by MessageToDict; `enforcing_*=0` is defense against Envoy treating an omitted consecutive value as 5. Helper stays on `Route`.
- **Serializer tests:** split is in the expected Meta.fields list; create test sets/asserts non-default split and consecutive.
- **CDS tests:** assert `consecutiveLocalOriginFailure` and `enforcingConsecutiveLocalOriginFailure` by key, not `.get(..., 0)`.

jay-steurer (approved):

PR #118, opened 2026-06-12
[AAP-76816] Add progress reporting during merge task execution

john-westcott-iv:
## Summary
- Add percentage-based progress logging to `migrate_service_data` at 5% threshold intervals
- Progress reported for both resource migration and role assignment migration phases
- Includes bookend messages at 0% (start) and 100% (completion)
- Uses `logger.info` so output appears in `podman logs` container output
- Service processing loop now shows index: `Processing service: controller (1/3)`

## Format
```
Migration progress [controller user role assignments]: 4300/86000 (5%)
Migration progress [controller user role assignments]: 8600/86000 (10%)
...
Migration progress [controller user role assignments]: 86000/86000 (100%)
```

## Test plan
- [ ] Unit tests for `_log_progress` threshold logic, edge cases (0 items, small counts, no duplicate 100%)
- [ ] DVT: run against cluster with 210+ seeded users, verify ~20 progress lines appear
- [ ] Verify progress appears in `podman logs` output

Depends on: #115 (AAP-76813)


aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1360,9 +1421,15 @@ def _fetch_role_assignments(self, assignment_actor: AssignmentActorType, service
             json_response = method(filters=filters).json()
             if total_count is None:
                 total_count = json_response.get('count', 0)
+                self._current_assignment_total = total_count
```
Dostonbek1, line 1426: **Observation (non-blocking):** `_current_assignment_total` is set here on the first page but is NOT updated when count drift occurs (the `elif` branch below updates local `total_count` but not `self._current_assignment_total`). This means:

- If items are **added** during migration, progress percentages can exceed 100% (e.g., `"87000/86000 (100%)"` followed by more processing with no further progress output)
- If items are **removed**, the 100% bookend may never fire since `processed` won't reach the original `total`

For a progress indicator this is acceptable behavior (and fixing it would require updating `_current_assignment_total` in the drift branch, which couples the generator further to the progress state). Just noting for awareness — if operators report confusing progress output, this is likely why.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -846,6 +899,12 @@ def migrate_resource(self, resource_type_name: str) -> None:
             "content_type__resource_type__name": resource_type_name,
         }
 
+        # Get initial total count for progress reporting
+        initial_count_response = self.client.list_resources(filters=api_call_filters).json()
```
Dostonbek1, line 905: **Minor note:** This issues an extra `list_resources` API call solely for the count. The very first iteration of the `while True` loop below will make the same call again (via `_get_filtered_resources`). For a long-running migration doing thousands of calls, one extra round-trip per resource type is negligible — but if you ever refactor `_get_filtered_resources` to return the full response (or just the count alongside results), you could eliminate this.

Dostonbek1 (commented):
## Code Review Summary

This PR adds operator-friendly progress reporting to the long-running `migrate_service_data` command, enabling monitoring via `podman logs`. The approach is well-designed: threshold-based logging at 5% intervals avoids flooding logs while giving visibility into progress. The service index addition (`Processing service: controller (1/3)`) is a nice touch.

**Files Reviewed**: 2 files
**Comments Posted**: 2 review comments

### Issues Found
- 0 security/safety issues
- 1 minor correctness observation (progress accuracy during count drift)
- 1 minor suggestion (extra API call)

### Overall Assessment
LGTM with minor observations. The `_log_progress` method is clean and well-tested, with good handling of edge cases (zero items, bookends, no-duplicate-100%). The threshold logic is correct.

### Highlights
- Clean separation of progress logic into `_log_progress` with clear docstring
- Good test coverage: threshold crossings, zero items, small counts, bookends, and no-duplicate-100%
- Using `logger.info` (not `self.stdout.write`) is the right choice for container log visibility
- `PROGRESS_STEP` as a class constant makes it easy to tune

### Note on PR scope
This PR includes commits from #115 (AAP-76813). My review focuses on the AAP-76816 progress-reporting changes specifically. I reviewed #115 separately.

Dostonbek1 (approved):

tznamena:
Added "DO NOT MERGE" label so I can use this PR to investigate why sonarcloud isn't invoked for it.

tznamena:
I removed the "do not merge" label because this branch has been rebased on top of "devel" and once the unit test pass, we should see sonarcloud code analysis run as well.

aap_gateway_api/management/commands/migrate_service_data.py
john-westcott-iv: Good catch. If items are added mid-pagination, we'd hit 100% early and keep processing silently. If removed, the 100% bookend won't fire (though the migration still completes). The drift warning from AAP-76813 already tells operators the count changed, so this is acceptable for a progress indicator. Will keep in mind if operators report confusing output.

aap_gateway_api/management/commands/migrate_service_data.py
john-westcott-iv: Agreed — one extra API call per resource type is negligible in a migration doing thousands of calls. Not worth coupling the progress tracking into _get_filtered_resources for now.

jeffh-oss (approved):

PR #120, opened 2026-06-16
[AAP-78674] Replace role assignment migration with set-diff approach

tznamena:
## Description

- **What is being changed?** The `migrate_role_assignments()` method in `migrate_service_data.py` is replaced with a set-diff approach. Instead of re-fetching and re-processing every upstream role assignment on each run, the command now builds local and remote assignment sets, computes `to_create = remote - local`, and only calls `give_permission` for the delta. Nine helper methods and the `AssignmentActorType` enum are removed in favor of four new methods and the shared `AssignmentTuple`/`get_local_assignments()` utilities from DAB.

- **Why is this change needed?** At scale (86K role assignments), the brute-force approach takes 4+ hours because every assignment triggers a `give_permission` call. On a reinstall where data is already synchronized, all 86K calls are no-ops — a complete waste. The current implementation also crashes with an uncaught `RuntimeError` when assignment counts change during pagination.

- **How does this change address the issue?** On a reinstall the delta is empty, so migration completes in seconds with zero `give_permission` calls. On a fresh install, behavior is identical to before. Count drift during pagination is handled with a two-pass recovery: if the first pass detects drift, a second pass runs on the delta only. Persistent drift raises `RuntimeError` so the installer can retry.

Requires: ansible/django-ansible-base#1024

## Type of Change

- [ ] Documentation update
- [x] Test update
- [x] Refactoring (no functional changes)
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

A running PostgreSQL instance matching the gateway test configuration, and the DAB submodule pointed at the PR 1024 branch (or devel once PR 1024 is merged).

### Steps to Test

1. Run new tests: `pytest aap_gateway_api/tests/management/test_migrate_role_assignments.py -v`
2. Run existing migration tests to verify no regressions: `pytest aap_gateway_api/tests/management/test_migrate_resources.py -v`
3. On a dev environment with synced data, run `python manage.py migrate_service_data --username admin` and observe:
   - Output includes `Set-diff: N remote, M local, 0 to create` for each service
   - Completes in seconds, not hours
4. On a fresh environment (no prior migration), run the same command and verify all role assignments are correctly created (same end state as before).

### Expected Results

- All new tests pass (20 tests covering `_build_assignment_tuple`, `_fetch_remote_assignment_set`, `migrate_role_assignments` set-diff logic, `_create_assignment_from_tuple`, and `_get_role_definitions_to_exclude`).
- All existing `test_migrate_resources.py` tests pass (tests for removed methods have been cleaned up).
- On a reinstall, zero `give_permission` calls are made and migration completes in seconds.
- On a fresh install, all role assignments are migrated correctly.

### Design notes

**Why not use `RemoteAssignmentFetcher` from the shared utils?**

1. It doesn't support the `not__role_definition__name__in` server-side filter needed for role exclusion across services (controller is authoritative for shared roles like Organization Admin).
2. It doesn't track count drift — only reports `is_complete` for HTTP errors.
3. It resolves `ansible_id_or_pk` as `object_ansible_id || object_id` for all types, but `get_local_assignments()` uses `object_ansible_id` only for org/team and raw `object_id` for everything else. This mismatch would cause false diffs on reinstalls.

**Key resolution consistency:** `_build_assignment_tuple` matches `get_local_assignments()` exactly — org/team content types use `object_ansible_id`, everything else uses the raw `object_id` from the service, and global assignments use `None`.

**Local set scoping:** `get_local_assignments()` is called with `service=None` (all assignments). This avoids the issue where org/team assignments have `content_type__service='shared'` rather than the migrating service name. Extra local assignments from other services are harmless since we only compute `to_create = remote - local`.

### DIFF
--- a/aap_gateway_api/management/commands/migrate_service_data.py
+++ b/aap_gateway_api/management/commands/migrate_service_data.py
@@ -1,12 +1,13 @@
 import logging
 from collections import OrderedDict
-from enum import Enum
-from typing import Any, Dict, Iterator, List, Optional, Tuple, Type
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Tuple, Type
 
 from ansible_base.authentication.models import AuthenticatorUser
 from ansible_base.authentication.models.authenticator import Authenticator
 from ansible_base.lib.utils.settings import get_setting
-from ansible_base.rbac.models import DABContentType, DABPermission, RoleDefinition, RoleTeamAssignment, RoleUserAssignment
+from ansible_base.rbac.assignment_utils import AssignmentTuple, get_local_assignments
+from ansible_base.rbac.models import DABContentType, DABPermission, RoleDefinition
 from ansible_base.rbac.remote import RemoteObject
 from ansible_base.resource_registry.constants import (
     SHARED_AAP_FLAG_RESOURCE_TYPE,
@@ -34,9 +35,10 @@ logger = logging.getLogger('aap.gateway.management.commands.migrate_service_data
 User = get_user_model()
 
 
-class AssignmentActorType(Enum):
-    TEAM = 'team'
-    USER = 'user'
+@dataclass
+class _RemoteFetchResult:
+    assignments: set = field(default_factory=set)
+    count_drifted: bool = False
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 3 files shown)


tznamena:
Closing and will re-open it against a feature branch.

PR #126, opened 2026-06-18
[AAP-70178] Reject invalid credentials for Hub container registry requests

jerabekjiri:
## Description
Requires: ansible/ansible.platform#214

Issue: [AAP-70178](https://redhat.atlassian.net/browse/AAP-70178)

Fixes podman login / docker login always reporting "Login Succeeded" even with invalid credentials.
 
When podman login sends invalid Basic credentials, the gateway catches the `AuthenticationFailed` exception but passes the request through to `pulp_container` as anonymous. The `/token/` endpoint then issues an anonymous token and podman sees a successful response on the `/v2/` retry — reporting "Login Succeeded" despite wrong credentials.

**What changed**
  - New `is_container_registry` flag on routes — a boolean field (default `False`) that marks routes serving container registry endpoints (`/v2/, /token/`). This will need to be set via the register_service command in the installers
  - 401 rejection for invalid credentials on registry routes — when `is_container_registry` is true and
  Basic auth fails, the gateway now returns a 401 immediately instead of falling through to anonymous. Anonymous requests (no credentials) still pass through normally for the token handshake.
  - Scoped to registry endpoints only — all other routes retain the existing pass-through behavior where failed auth falls back to anonymous.

## Type of Change

- [ ] Documentation update
- [ ] Test update
- [ ] Configuration change

## Self-Review Checklist

### Steps to Test
1. Run AAP
2. 
3. 

### Required Actions

- [ ] Blocked by PR/MR: #XXX
  

### Screenshots/Logs

[AAP-70178]: https://redhat.atlassian.net/browse/AAP-70178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Introduced a per-route setting to reject failed/invalid Basic credentials with an HTTP 401, instead of allowing the request to continue to the backend.
  * Enabled this behavior for container registry token issuance (for `/token/`) while keeping normal registry access flow for `/v2/`.

* **Bug Fixes**
  * Invalid Basic credentials can now be blocked earlier when the setting is enabled, with the response indicating “Invalid credentials”.

* **Tests**
  * Added coverage for valid/invalid authentication and the route-specific rejection behavior.


aap_gateway_api/proxy/control_plane.py
```diff
@@ -122,6 +123,25 @@ def _is_oidc_public_path(self):
...
+        anonymous token and clients like podman report login success despite
+        invalid credentials. For registry requests, the gateway must reject
+        invalid credentials immediately.
+        """
+        if self.service_type != DefaultServiceType.HUB.value:
+            return False
+        container_registry_prefixes = (
+            '/v2',
+            '/token',
+        )
```
john-westcott-iv: I'm not sure we want to hard code these paths in here vs having something on the route to indicate that is a registry. Imagine if controller wanted to expose a registry, they would have to update this portion of the code instead of just being able to add a route.

jerabekjiri: hi, thanks for the feedback. Could you elaborate more on what would be the right approach here? Do you mean to add a boolean flag to the `Route` mode with something like `is_container_registry` or am I misunderstanding your comment?

john-westcott-iv: Yea, I would think something like an is_container_registry bool on the route object would suit nicely here. This would, of course, mean installer changes as well (to set the flag on route creation). But then your code here would just be `if route.is_container_registry():`

jerabekjiri: done :+1: now the route has attribute `is_container_registry`

aap_gateway_api/proxy/control_plane.py
```diff
@@ -233,8 +236,10 @@ def try_authenticate_request(self):
         try:
             user = self.drf_request.user
         except AuthenticationFailed:
-            # Rest framework will raise this exception if the user/pass combo is invalid.
-            # If this is the case we want to fall though and _return_not_authenticated so that the Authorization header will be sent to the backend.
+            # Reject immediately for container registry requests with invalid credentials,
+            # otherwise the token endpoint silently issues an anonymous token and podman reports login success.
+            if self.is_container_registry:
+                return self._return_no_auth_with_reason("Invalid credentials.", http_status_code=401, code=16)
```
Dostonbek1, line 243: **Correctness**: The condition here rejects *all* `AuthenticationFailed` exceptions on container registry routes, but the bug being fixed is specifically about **Basic auth** credentials. If a registry route receives a failed Bearer token or other auth scheme, it should still fall through to anonymous (or be handled by its own logic).

This was also flagged by CodeRabbit — I agree with their suggestion to scope the check to Basic auth:

```suggestion
            auth_header = self.drf_request.META.get("HTTP_AUTHORIZATION", "")
            if self.is_container_registry and auth_header.startswith("Basic "):
                return self._return_no_auth_with_reason("Invalid credentials.", http_status_code=401, code=16)
```

This ensures that only explicitly provided *invalid Basic credentials* are rejected, while other auth failures on registry routes still pass through as before.

aap_gateway_api/tests/proxy/test_proxy.py
```diff
@@ -241,6 +244,81 @@ def test_auth_header_selection(self, ext_auth, service_type, auth_type, expected
...
+        )
+
+        with mock.patch(
+            "aap_gateway_api.authentication.service_token_auth.ServiceTokenAuthentication.authenticate",
+            side_effect=AuthenticationFailed("Invalid credentials"),
+        ):
+            response = ext_auth.Check(request, None)
+
+        assert response.status.code == 0
+
```
Dostonbek1, line 337: **Test gap**: Consider adding a test case for a non-Basic auth failure on a registry route (e.g., a `Bearer` token that triggers `AuthenticationFailed`). This would verify that the new denial is scoped only to Basic auth and doesn't accidentally block other auth methods on registry endpoints.

Something like:

```python
def test_non_basic_auth_failure_passes_through_for_registry_route(self, ext_auth):
    """Non-Basic auth failure on a registry route should pass through (not be rejected)."""
    request = Request(
        path="/token/",
        service_type="hub",
        is_container_registry="t",
        header_diff={"AUTHORIZATION": "Bearer some-invalid-token"},
    )

    with mock.patch(
        "aap_gateway_api.authentication.service_token_auth.ServiceTokenAuthentication.authenticate",
        side_effect=AuthenticationFailed("Invalid token"),
    ):
        response = ext_auth.Check(request, None)

    assert response.status.code == 0
```

aap_gateway_api/tests/fixtures/service_config.yml
```diff
@@ -27,6 +27,10 @@ services:
       - name: "hub container registry"
         description: "Hub Container Registry."
         gateway_path: "/v2/"
+      - name: "hub container registry tokens"
+        description: "Hub Container Registry Tokens."
+        gateway_path: "/token/"
```
Dostonbek1, line 32: **Clarification question**: The `/v2/` route above is labeled "hub container registry" but does not have `is_container_registry: true`, while `/token/` does. This is likely intentional (the vulnerability is in the token endpoint flow), but it would be helpful to add a comment here explaining *why* only `/token/` is marked — e.g.:

```yaml
# Only /token/ is marked as container registry because that's where
# the anonymous token issuance happens; /v2/ relies on the standard
# Www-Authenticate challenge flow.
```

Alternatively, consider whether `/v2/` should also have the flag set for defense-in-depth, since invalid Basic creds sent directly to `/v2/` would still fall through as anonymous today.

Dostonbek1 (commented):
## Code Review Summary

This PR fixes a real security-relevant bug where podman/docker login always reports "Login Succeeded" even with invalid credentials. The approach is sound — adding an `is_container_registry` flag to routes and rejecting invalid credentials at the gateway level before they fall through to anonymous.

**Files Reviewed**: 8 files
**Comments Posted**: 3 review comments

### Issues Found
- 1 correctness/logic issue (condition too broad)
- 1 code quality suggestion (test fixture clarity)
- 1 test coverage gap

### Overall Assessment
Needs minor fixes — specifically, the denial condition should be scoped to Basic auth only to avoid rejecting other auth schemes on registry routes.

### General Feedback
- Good test coverage for the core scenarios (invalid, valid, anonymous, non-registry fallback)
- Clean integration with the existing route model and control plane architecture
- The migration and serializer changes are straightforward

Dostonbek1:
@jerabekjiri Could you address the above review comments and re-request a new review once updated?

aap_gateway_api/models/route.py
```diff
@@ -67,6 +67,11 @@ class Meta:
     is_internal_route = models.BooleanField(
         default=False, help_text=_("If true, the gateway will only allow other Ansible services to access this route. Requires gateway auth to be enabled.")
     )
+    # Indicates that this route serves a container registry (e.g. /v2/, /token/).
+    is_container_registry = models.BooleanField(
+        default=False,
+        help_text=_("If true, this route serves a container registry endpoint and the gateway will enforce Basic credential validation during the auth flow."),
```
john-westcott-iv: Can we reword this to not use "gateway"? I.e. "if true this route will return a 401 instead of allowing unauthenticated requests to the back end".

I'm also wondering if a container registry is the only reason we would want to do this? I.e. should we just call this something like prevent_passthrough_authenticaton or something along those lines?

john-westcott-iv (commented):
For the test completeness we will need a pr against the ansible.platform repo and then reference that in the PR body like "Requires: ansible/ansible.platform#XXX". That will make CI run with that version of the collection.

aap_gateway_api/proxy/control_plane.py
jerabekjiri: done 👍

aap_gateway_api/tests/fixtures/service_config.yml
jerabekjiri: comment added

aap_gateway_api/tests/proxy/test_proxy.py
jerabekjiri: test added, thanks for suggestion :+1:

aap_gateway_api/models/route.py
jerabekjiri: I like the name change suggestion and make it more general. Updated to `reject_failed_auth` which I think might be a better fit, but I'm open to your suggestion, wdyt?

tznamena:
@jerabekjiri One of CodeRabbit's comments is still unaddressed — the `help_text` for `reject_failed_auth` (in both the model and the migration) says *"return a 401 instead of allowing unauthenticated requests to the back end"*, but the actual behavior only rejects requests with **invalid Basic credentials**. Truly anonymous requests (no credentials at all) still pass through. This is misleading for administrators. 

Suggested wording: *"If true, requests with invalid Basic credentials will receive a 401 instead of being passed to the back end as anonymous."*

Please update both `aap_gateway_api/models/route.py` and `aap_gateway_api/migrations/0024_route_reject_failed_auth.py`.

aap_gateway_api/proxy/control_plane.py
```diff
@@ -286,6 +293,7 @@ def Check(self, request, context):
 
         self.request_path = request.attributes.request.http.path
         self.is_internal_route = self.is_route_internal(request)
+        self.reject_failed_auth = self.reject_failed_auth(request)
```
tznamena: **Bug: method shadows itself**

`self.reject_failed_auth` is a method defined on `_ExternalAuth` (line 87). This assignment calls the method and then overwrites it with the boolean result on the instance. While it happens to work because `Check()` runs once per request, it's fragile and inconsistent with the pattern directly above:

```python
self.is_internal_route = self.is_route_internal(request)  # different names ✓
self.reject_failed_auth = self.reject_failed_auth(request)  # shadows the method ✗
```

Please either:
- Rename the method (e.g., `should_reject_failed_auth`) to match the `is_route_internal` / `self.is_internal_route` naming pattern, or
- Use a different attribute name (e.g., `self._reject_failed_auth`) and update all references in `try_authenticate_request`.

tznamena (approved):
All review comments from all reviewers have been addressed:

1. **`@pytest.mark.skip` removed** — tests now run without skip markers
2. **401 rejection scoped to Basic auth only** — checks `auth_header.startswith("Basic ")` before rejecting
3. **Method self-shadowing fixed** — method renamed to `should_reject_failed_auth()`, no longer collides with the boolean attribute
4. **Field renamed from `is_container_registry` to `reject_failed_auth`** — more general and reusable
5. **Help text updated** — accurately describes "invalid Basic credentials" behavior, no "gateway" wording
6. **Non-Basic auth test added** — `test_non_basic_auth_failure_passes_through_for_registry_route` confirms Bearer failures pass through
7. **Comment added to `service_config.yml`** — explains why `/token/` has the flag but `/v2/` does not

LGTM!

aap_gateway_api/models/route.py
```diff
@@ -67,6 +67,13 @@ class Meta:
     is_internal_route = models.BooleanField(
         default=False, help_text=_("If true, the gateway will only allow other Ansible services to access this route. Requires gateway auth to be enabled.")
     )
+    reject_failed_auth = models.BooleanField(
+        default=False,
+        help_text=_(
+            "If true, requests with invalid Basic credentials will receive a 401 instead of being passed to the back end"
+            " as anonymous. Requests without credentials continue to be passed through anonymously."
```
john-westcott-iv: Is this actually correct? I thought if gateway failed authentication we just forwarded the request "as-is" to the backend whether anonymous or not?

aap_gateway_api/models/route.py
```diff
@@ -67,6 +67,13 @@ class Meta:
     is_internal_route = models.BooleanField(
         default=False, help_text=_("If true, the gateway will only allow other Ansible services to access this route. Requires gateway auth to be enabled.")
     )
+    reject_failed_auth = models.BooleanField(
```
john-westcott-iv: Since we're limiting this to Basic auth only do we want to make this "reject_failed_basic_auth"?

aap_gateway_api/proxy/control_plane.py
```diff
@@ -233,8 +236,12 @@ def try_authenticate_request(self):
         try:
             user = self.drf_request.user
         except AuthenticationFailed:
-            # Rest framework will raise this exception if the user/pass combo is invalid.
-            # If this is the case we want to fall though and _return_not_authenticated so that the Authorization header will be sent to the backend.
+            # Reject immediately for container registry requests with invalid credentials,
+            # otherwise the token endpoint silently issues an anonymous token and podman reports login success.
+            auth_header = self.drf_request.META.get("HTTP_AUTHORIZATION", "")
```
john-westcott-iv, line 241: This check is case-sensitive but HTTP auth schemes are case-insensitive per RFC 7235. DRF's own `BasicAuthentication` checks case-insensitively (`auth[0].lower() != b'basic'`), so a client sending `"basic user:pass"` would trigger `AuthenticationFailed` but bypass this rejection (falling through to anonymous).

All standard OCI clients send `"Basic"` with capital B, so real-world risk is low, but for correctness:

```suggestion
            if self.reject_failed_auth and auth_header[:6].lower() == "basic ":
```

john-westcott-iv (changes requested):
Looks good overall — all prior review concerns have been addressed and CI is passing. A few items to address before merge. See inline comments.

aap_gateway_api/models/route.py
jerabekjiri: oh, that's true! I got confused by coderabbit and only thought about the case for Hub, but the field isn't specific to that anymore.

aap_gateway_api/models/route.py
jerabekjiri: why didn't I think of that :+1:

aap_gateway_api/models/route.py
jerabekjiri: updated to the `If true, requests with invalid Basic credentials will receive a 401 instead of being passed to the back end. Requests without credentials continue to be passed through anonymously.`

I think the second sentence is still valid though?

john-westcott-iv: Yes, that should still be correct. Thanks!

john-westcott-iv (approved):
All prior review comments addressed:
- Renamed to `reject_failed_basic_auth` — accurately reflects the scope
- Help text reworded — no longer references "gateway"
- Case-insensitive Basic auth check applied (`auth_header[:6].lower()`)

CI passing. LGTM.

PR #131, opened 2026-06-19
[AAP-76816] Add progress reporting during merge task execution

john-westcott-iv:
## Summary
- Add percentage-based progress logging to `migrate_service_data` at 5% threshold intervals
- Progress reported for both resource migration and role assignment migration phases
- Includes bookend messages at start and 100% completion
- Uses `logger.info` so output appears in `podman logs` / container log output
- Service processing loop now shows index: `Processing service: controller (1/3)`

## Format
```
Migration progress [controller user role assignments]: 4300/86000 (5%)
Migration progress [controller user role assignments]: 8600/86000 (10%)
...
Migration progress [controller user role assignments]: 86000/86000 (100%)
```

## Test plan
- [ ] Unit tests for `_log_progress` threshold logic, edge cases (0 items, small counts, no duplicate 100%)
- [ ] DVT: run against cluster with 210+ seeded users, verify ~20 progress lines appear
- [ ] Verify progress appears in `podman logs` output

Replaces: #118 (closed due to target branch deletion)

## Summary by CodeRabbit

* **New Features**
  * Enhanced migration visibility with percentage-based progress logging for services, resources, and role assignments.
  * Service processing output now includes the current service index/total.
* **Bug Fixes**
  * Progress reporting correctly handles zero-item scenarios and avoids duplicate “100%” logs when thresholds are hit exactly.
* **Tests**
  * Added targeted coverage for progress logging threshold behavior, including zero items, small-count skipping, and consistent bookend messages.

### DIFF
--- a/aap_gateway_api/management/commands/migrate_service_data.py
+++ b/aap_gateway_api/management/commands/migrate_service_data.py
@@ -64,6 +64,9 @@ class Command(BaseCommand):
     RESOURCE_DATA_FILTERS = {"extra_fields": "resource_data"}
     BIG_PAGE_FILTERS = {"page_size": str(get_setting('RESOURCE_LIST_MAX_PAGE_SIZE', DEFAULT_MAX_PAGE_SIZE))}
 
+    # Progress reporting step size (percentage)
+    PROGRESS_STEP = 5
+
     # Service processing order - Controller first to establish priority for user merging
     SERVICE_TYPE_ORDER = [
         DefaultServiceType.CONTROLLER.value,
@@ -75,6 +78,37 @@ class Command(BaseCommand):
 
     There is no option to control merging of users, because users are never migrated.
     The exception is that the provided --username, which will be merged."""
+
+    def _log_progress(self, label: str, processed: int, total: int) -> None:
+        """
+        Log migration progress when a percentage threshold is crossed.
+
+        Emits a log line at every PROGRESS_STEP% interval. Always logs a starting
+        message on the first item and a 100% message on the last item. If a single
+        item crosses multiple thresholds, only the highest is reported.
+        """
+        if total == 0:
+            logger.info("Migration progress [%s]: 0 items to process", label)
+            return
+
+        percent = (processed / total) * 100
+        threshold = (int(percent) // self.PROGRESS_STEP) * self.PROGRESS_STEP
+        last = self._progress_thresholds.get(label, -1)
+
+        if processed == total:
+            if last < 100:
+                self._progress_thresholds[label] = 100
+                logger.info("Migration progress [%s]: %d/%d (100%%)", label, processed, total)
+            return
+
+        if processed == 1 and last < 0:
+            self._progress_thresholds[label] = threshold
+            logger.info("Migration progress [%s]: %d/%d (%d%%)", label, processed, total, threshold)
+            return
+
+        if threshold > last:
+            self._progress_thresholds[label] = threshold
+            logger.info("Migration progress [%s]: %d/%d (%d%%)", label, processed, total, threshold)
 
     def add_arguments(self, parser) -> None:
         """
@@ -214,6 +248,7 @@ class Command(BaseCommand):
             )
 
         self._services_with_count_drift = set()
+        self._progress_thresholds = {}
 
         # Merge all partially migrated users before proceeding with migration
         self.stdout.write("\n=== Merging partially migrated users ===")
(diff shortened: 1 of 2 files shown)


aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -871,6 +907,12 @@ def migrate_resource(self, resource_type_name: str) -> None:
             "content_type__resource_type__name": resource_type_name,
         }
 
+        # Get initial total count for progress reporting
+        initial_count_response = self.client.list_resources(filters=api_call_filters).json()
+        resource_total = initial_count_response.get('count', 0)
```
tznamena: These two lines are duplicating the API call that's done as first thing in the `while True` loop. Can we initialize the total by setting it to `None` here and updating it at the start of the loop? Would save one API call.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1414,10 +1459,15 @@ def migrate_role_assignments(self, assignment_actor: AssignmentActorType, servic
         4. Giving permission in gateway to the actor for the object (handling both Resources and RemoteObjects)
         """
 
+        progress_label = f"{service_slug} {assignment_actor.value} role assignments"
+        self._current_assignment_total = 0
```
tznamena, line 1465: A comment would be helpful here to explain that the value is set from within `_fetch_role_assignments` on the first iteration.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -871,6 +907,12 @@ def migrate_resource(self, resource_type_name: str) -> None:
             "content_type__resource_type__name": resource_type_name,
         }
 
+        # Get initial total count for progress reporting
+        initial_count_response = self.client.list_resources(filters=api_call_filters).json()
+        resource_total = initial_count_response.get('count', 0)
+        resource_processed = 0
+        progress_label = f"{resource_type_name} resources"
```
tznamena: Shouldn't this label include service as well?

aap_gateway_api/tests/management/test_migrate_resources.py
```diff
@@ -2438,6 +2443,7 @@ def test_migrate_role_assignments_catches_give_permission_error(admin_user, caps
     """Test that errors in give_permission are caught separately from fetch errors."""
     cmd = MigrateCommand()
     cmd._services_with_count_drift = set()
+    cmd._progress_thresholds = {}
```
tznamena, line 2446: This line is repeated several times, could use a helper to avoid duplicating it.

aap_gateway_api/tests/management/test_migrate_resources.py
```diff
@@ -2581,3 +2587,84 @@ def test_give_permission_failure_does_not_block_migration(admin_user, capsys, se
...
+        cmd._log_progress("bookend", 7, 7)
+
+    progress_msgs = [msg for msg in caplog.messages if "Migration progress [bookend]" in msg]
+    assert len(progress_msgs) >= 2
+    assert "1/7" in progress_msgs[0]
+    assert "7/7" in progress_msgs[-1]
+    assert "(100%)" in progress_msgs[-1]
+
+
+@pytest.mark.django_db
```
tznamena: Is this marker needed for testing `_log_progress()`?

tznamena (commented):
A few small comments and questions. Looks good otherwise.

aap_gateway_api/management/commands/migrate_service_data.py
john-westcott-iv: Good catch — eliminated the extra call. _get_filtered_resources now returns (results, count) and the total is captured from the first loop iteration.

aap_gateway_api/management/commands/migrate_service_data.py
john-westcott-iv: Added a comment explaining the side-effect.

aap_gateway_api/management/commands/migrate_service_data.py
john-westcott-iv: The role assignment label already includes the service slug: `f"{service_slug} {assignment_actor.value} role assignments"`. This was set correctly from the start — only the resource label was missing it, which was fixed in the previous commit.

aap_gateway_api/tests/management/test_migrate_resources.py
john-westcott-iv: It's a single-line init (`cmd._progress_thresholds = {}`). A helper would add more code than it saves and reduce readability by hiding a straightforward assignment.

aap_gateway_api/tests/management/test_migrate_resources.py
john-westcott-iv: You're right — _log_progress is pure Python with no DB access. Removed the markers.

aap_gateway_api/tests/management/test_migrate_resources.py
tznamena: Fair point.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: OK

tznamena (approved):
LGTM

PR #132, opened 2026-06-19
[AAP-78674] Replace role assignment migration with PK cursor approach

tznamena:
## Description

- **What is being changed?** The role assignment migration in `migrate_service_data.py` is replaced with a PK cursor approach. Instead of re-fetching and calling `give_permission` for every upstream assignment on each run, the command now stores the last-processed assignment PK per service and queries with `id__gt=<last_pk>&order_by=id` to fetch only new assignments. The old brute-force methods (`_fetch_role_assignments`, `migrate_role_assignments`, `_resolve_role_definition`, `_resolve_gateway_actor`, `_resolve_content_object`, and related helpers) are replaced by cursor-based pagination with per-page persistence. A raw-SQL `_CursorStore` class manages the cursor (no Django model, no migration file).

- **Why is this change needed?** At scale (100K+ role assignments), the brute-force approach takes 4+ hours because every assignment triggers a `give_permission` call. On a reinstall where data is already synchronized, all 100K+ calls are no-ops. This blocks the installer for hours doing zero useful work.

- **How does this change address the issue?** On a reinstall the cursor is past the last PK, so the API returns zero results and migration completes immediately — no local set query, no remote fetch, O(1) memory. On a fresh install, all assignments are fetched and created with the cursor advancing per page. HTTP errors raise RuntimeError immediately (no per-page retry) — the PK cursor provides crash recovery, so the installer re-runs the command and the cursor resumes from the last completed page. Cursor persistence means a crash loses at most one page of work.

No external dependencies. The `AssignmentTuple` dependency from django-ansible-base PR #1024 has been removed — assignment creation now works directly with raw API response dicts.

Requires: ansible/django-ansible-base/pull/1032

## Type of Change

- [ ] Documentation update
- [x] Test update
- [x] Refactoring (no functional changes)
- [ ] Configuration change

## Self-Review Checklist

### Prerequisites

A running PostgreSQL instance matching the gateway test configuration.

### Steps to Test

1. Run new unit tests: `pytest aap_gateway_api/tests/management/test_migrate_service_data_role_assignments.py -v`
2. Run existing migration tests: `pytest aap_gateway_api/tests/management/test_migrate_resources.py -v`
3. On a dev environment, run `python manage.py migrate_service_data --username admin`:
   - First run: all assignments created, cursor rows populated in DB
   - Second run: cursor past last PK, zero results, completes immediately
4. Verify `migrate_service_data_role_cursor` table has correct entries per service/type
5. Simulate crash (Ctrl+C mid-run), re-run, verify it resumes from cursor

### Expected Results

- 33 new unit tests pass (cursor store, cursor pagination, orchestration, assignment creation, role exclusion)
- All existing `test_migrate_resources.py` tests pass
- The reinstall integration test confirms zero work on second run
- On a reinstall, zero API results and "0 assignments created" in output

### Required Actions

- [x] No downstream repository changes required
  - The `AssignmentTuple` dependency from django-ansible-base PR #1024 has been removed. The `_CursorStore` uses raw SQL and requires no Django migration.

### Error handling and exit codes

- **Clean run**: all assignments created, flag set, exit 0
- **Reinstall (cursor past end)**: 0 results, flag set, exit 0
- **HTTP error**: RuntimeError immediately (no per-page retry) → service fails → CommandError → non-zero exit → installer/reconcile retries → cursor resumes from last completed page
- **Crash mid-page**: cursor at last complete page, at most 1 page replayed on re-run (idempotent via get_or_create)
- **DB error in cursor**: graceful degradation to last_pk=0, reprocesses all assignments (safe because idempotent)

### Performance at scale

| Scenario | Before (brute-force) | After (PK cursor) |
|----------|---------------------|-------------------|
| Reinstall (0 new) | 4+ hours, 100K+ give_permission calls | < 1 second, 1 API call |
| 300 new items | 4+ hours (fetches all 100K+) | < 10 seconds |
| Memory | O(100K+) assignment objects | O(page_size) constant |
| Crash at 50% | Restart from scratch | Resume from cursor |

### Why not use `RemoteAssignmentFetcher` from the shared utils?

1. It doesn't support the `not__role_definition__name__in` server-side filter needed for role exclusion across services
2. It doesn't support cursor-based pagination (`id__gt` filtering)
3. Its key resolution (`object_ansible_id || object_id` for all types) doesn't match the gateway's content type branching in `_create_assignment`


jeffh-oss (commented):
## Code Review: PR #132 — [AAP-78674] Replace role assignment migration with set-diff approach

### Verdict: READY_FOR_HUMAN_REVIEW

### Scores
| Lens | Score | Findings |
|---|---|---|
| Functionality | 10/10 | 0 deductions |
| Security | 10/10 | 1 Nit (0 pts) |
| Quality | 10/10 | 1 Nit (0 pts) |
| **Overall** | **10.0/10** | |

### Findings
#### Critical
None.

#### Major
None.

#### Minor
None.

#### Nit

1. **[Quality]** **[migrate_service_data.py:1389]** Dense walrus-operator comprehension reduces readability
   - **Evidence**: `result.assignments.update(t for a in data.get('results') or [] if (t := self._build_assignment_tuple(a, assignment_type)) is not None)`
   - **Confidence**: MEDIUM
   - **Fix**: Style preference — could be expanded to a for-loop for readability, but functionally correct as-is.
   - **Points**: 0

2. **[Security]** **[migrate_service_data.py:1410]** Content-type model extraction uses `split('.')[-1]` which always returns a non-empty string for non-empty input, including malformed values. The old code had explicit `ValueError` handling for malformed `content_type` values. In the new code, a malformed content type falls through to the `object_id` branch, producing a valid `AssignmentTuple`. The downstream `_create_assignment_from_tuple` handles errors gracefully, so no incorrect permissions would be granted.
   - **Evidence**: `model = content_type_str.split('.')[-1] if content_type_str else ''`
   - **Confidence**: LOW
   - **Fix**: Not required — downstream error handling is sufficient. Defense-in-depth improvement only.
   - **Points**: 0

### Needs Human Judgment

- **`AssignmentTuple` and `get_local_assignments()` contract**: The PR relies on an external dependency (`ansible_base.rbac.role_sync_utils` from django-ansible-base PR #1024). The key resolution logic in `_build_assignment_tuple` must match `get_local_assignments()` exactly — org/team content types use `object_ansible_id`, everything else uses `object_id`. A mismatch would cause false diffs on reinstalls (phantom creates). The reviewer should verify that DAB PR #1024 is merged and the contract is stable.

- **Branching mismatch between tuple builder and creator**: `_build_assignment_tuple` branches on the remote API's `content_type` string model, while `_create_assignment_from_tuple` branches on the local `role_definition.content_type.model`. These should always agree for legitimate assignments, but the reviewer should confirm this invariant holds across all service types (controller, hub, eda).

### Verification Results
| Stale refs: AssignmentActorType | `grep -rn "AssignmentActorType" aap_gateway_api/ --include="*.py"` | (no output) | PASS (0 matches) |
| Stale refs: deleted methods | `grep -rn "_resolve_role_definition\|_resolve_gateway_actor\|_resolve_content_object\|_format_*\|_SKIP" aap_gateway_api/ --include="*.py"` | (no output) | PASS (0 matches) |
| Stale refs: _fetch_role_assignments | `grep -rn "_fetch_role_assignments" aap_gateway_api/ --include="*.py"` | (no output) | PASS (0 matches) |
| Removed imports still needed | `grep -rn "RoleUserAssignment\|RoleTeamAssignment" aap_gateway_api/management/commands/migrate_service_data.py` | (no output) | PASS (not used) |

### Observations

1. **HTTP status code validation improvement**: The new code adds explicit `response.status_code != 200` checking before parsing JSON, which the old code lacked. This is a security and reliability improvement.
2. **Count drift logging**: The new `_paginate_assignments` does not update `initial_count` after detecting drift (unlike the old code). Subsequent pages will continue logging drift warnings against the original count. Functionally harmless — design choice, not a defect.
3. **Test coverage is thorough**: 35 new tests covering all code paths — tuple building, remote fetching, set-diff logic, creation error handling, drift exit code, role exclusion, and a reinstall-is-noop integration test.
4. **Mock pattern migration**: All 6 instances of the old `.return_value.json.return_value` chaining pattern replaced with proper Mock objects with `status_code=200`. The `_setup_empty_assignment_mocks` helper is well-documented and consistently reused.

### Limitations
This review was performed by an AI agent. It does not understand business context, domain intent, organizational constraints, or deployment environment specifics. LOW-confidence findings and items in "Needs Human Judgment" require human verification. This review is a first pass, not a final approval.

### Validation Summary
| Functionality | 0 | 0 | 0 | 0 |
| Security | 1 | 0 | 0 | 1 (Nit, 0 pts) |
| Quality | 1 | 0 | 0 | 1 (Nit, 0 pts) |

### Final Score
**Overall: 10.0/10 — READY_FOR_HUMAN_REVIEW**

---

jeffh-oss (approved):

john-westcott-iv (changes requested):
## Code Review: [AAP-78674] Replace role assignment migration with set-diff approach

**Score: 9.2/10 — NEEDS_CHANGES**

| Lens | Score |
|---|---|
| Functionality | 9/10 |
| Security | 10/10 |
| Quality | 8.5/10 |

### Major — Post-creation drift detection gap

The switch from an interleaved generator to a snapshot-based set-diff widens the drift detection blind spot. In the old code, pagination and processing were interleaved (`yield` per page), so new items appearing on later pages were naturally fetched and processed. In the new code, ALL pages are fetched into a set before ANY creation happens.

**The problem:** Items added to the remote system *after* pagination completes but *during* the creation phase are silently missed — `count_drifted` stays `False`, the migration flag is set, and the command exits 0. At the customer's scale (~220K assignments, ~300 new items per 10-minute window), if pagination takes ~10 minutes and creation takes ~10 minutes, approximately 300 assignments could be missed per run with no indication of a problem.

**Suggested fix — page-at-a-time set-diff:** Instead of fetching all remote assignments into memory then diffing, process the set-diff page by page:

1. Load `local_set = get_local_assignments()` once
2. For each page of remote assignments:
   - Build tuples from the page
   - `page_to_create = page_tuples - local_set`
   - Create only the missing ones
   - Check count drift per page (same as old code)

This preserves the O(delta) performance win (only create what's missing) while keeping the narrow per-page drift detection window from the old code. It also halves the memory footprint (no need to hold the full remote set).

### Minor findings

1. **DRY:** The second pass duplicates the first pass fetch-diff-create pattern (~10 lines). Consider extracting a helper.
2. **Missing test:** No test for mid-pagination HTTP error (page 1 succeeds, page 2 returns 500) — existing test only covers first-page failure.
3. **Test file naming:** Per conventions, `test_migrate_role_assignments.py` should be consolidated into `test_migrate_service_data.py`.

### Nits

1. **"using set-diff" / "Set-diff:" in stdout messages** — implementation detail meaningless to operators. Suggest: `"Migrating role assignments from {service_slug} (type {service_type_name})"` and `"Role assignments: {N} remote, {M} local, {K} to create"`.
2. **Side-effecting generator predicate:** `sum(1 for t in to_create if self._create_assignment_from_tuple(t))` hides the actual work inside an `if` clause. An explicit `for` loop with a counter would be clearer.
3. **Type annotation:** `_RemoteFetchResult.assignments: set` → `set[AssignmentTuple]`.

### Incidental (pre-existing)

- `"=== Re-enabling service authentication ==="` and `"Service authentication is now enabled"` imply a state transition the operator never witnessed — there's no corresponding "disabled" message. Consider rewording.

### What's good

- The set-diff approach is architecturally sound and the performance win is massive (100-500x for reinstall/partial sync)
- HTTP status code checking before `.json()` is a nice improvement
- `CommandError` on drift producing non-zero exit for installer retry is correct
- Test coverage is thorough — 35+ new tests covering tuple building, pagination, drift, and error handling
- The E2E reinstall no-op test directly validates the core performance scenario

Full review output available on request.

---

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+
+        The cursor is advanced after each fully-processed page, providing
+        crash safety — at most one page of work is lost on a crash or kill.
         """
-        role_definitions_to_exclude = self._get_role_definitions_to_exclude(service_type_name)
         page = 1
-        total_count = None  # we will check this on each page to see if anything changed
+        created = 0
+
         while True:
```
john-westcott-iv, line 1539: ### Critical: Page counter + cursor advancement skips items on fresh install

`_paginate_and_create` advances the PK cursor (`id__gt`) after each page while also incrementing the `page` counter. The upstream API uses offset-based pagination, so combining a changing `id__gt` filter with an incrementing page number causes items to be silently skipped.

**Example** (page_size=100, 1000 assignments, fresh install):

| Iter | page | id\_\_gt | Filtered set | Returns | Skipped |
|------|------|---------|--------------|---------|----------|
| 1 | 1 | (none) | 1000 items | 1-100 | -- |
| 2 | 2 | 100 | 900 items | 201-300 | **101-200** |
| 3 | 3 | 300 | 700 items | 501-600 | **301-500** |

Result: ~400 of 1000 processed. 600 permanently skipped (cursor advances past them).

**Why tests don't catch it:** E2E tests use <10 assignments (one page). The multi-page unit test mocks `list_fn` with `side_effect` that returns canned responses regardless of actual filter values.

**Fix:** Snapshot the cursor at start of run. Use it as a static `id__gt` filter for all pages. Still advance the cursor in the DB per page for crash recovery, but don't mutate the HTTP filter mid-run.

This method should also absorb:
- `_build_page_filters` — 2/3 of its args are static; build the filters inline
- `_extract_last_pk` — one-liner: `results[-1].get('id')`
- `_fetch_page_with_retry` / `HTTP_RETRY_LIMIT` — the command is designed to be re-run; per-page retry is redundant with the cursor + re-run design

**Suggested replacement:**

```python
def _paginate_and_create(self, list_fn, assignment_type, roles_to_exclude, cursor):
    base_filters = {'order_by': 'id', **self.BIG_PAGE_FILTERS}
    if roles_to_exclude:
        base_filters['not__role_definition__name__in'] = ','.join(roles_to_exclude)
    if cursor.last_pk > 0:
        base_filters['id__gt'] = str(cursor.last_pk)

    page = 1
    created = 0
    while True:
        response = list_fn(filters={**base_filters, 'page': page})
        if response.status_code != 200:
            raise RuntimeError(
                f"Failed to fetch {assignment_type} assignments "
                f"page {page}: HTTP {response.status_code}"
            )
        data = response.json()
        results = data.get('results') or []
        if not results:
            break
        created += self._create_page_assignments(results, assignment_type)
        last_id = results[-1].get('id')
        if last_id is not None:
            cursor.advance(last_id)  # DB only, does NOT update cursor.last_pk
        if not data.get('next'):
            break
        page += 1
    return created
```

Where `cursor` is a `_CursorStore` instance (see comment on migrate_data.py) whose `advance()` only persists to the DB -- it does NOT update `self.last_pk`, keeping the HTTP filter immutable for the run.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+    def _extract_last_pk(results: List[Dict[str, Any]]) -> Optional[int]:
+        """Return the id of the last item in a page of results, or None."""
+        for item in reversed(results):
+            pk = item.get('id')
+            if pk is not None:
+                return pk
+        return None
 
-        It performs this migration by:
+    @staticmethod
```
john-westcott-iv: Consider whether the `AssignmentTuple` indirection is justified. The old code resolved role definitions, actors, and content objects inline during iteration -- more direct, fewer abstractions.

The tuple exists to match DAB's `role_sync_utils.AssignmentTuple` for set-diff comparison, but since this PR uses a PK cursor (not set-diff), the tuple is just an intermediate data structure adding a layer between "parse API response" and "call give_permission". If the DAB dependency on `AssignmentTuple` isn't load-bearing, the original inline approach is simpler.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+            if assignment_tuple.ansible_id_or_pk is None:
+                role_definition.give_global_permission(actor)
+            elif role_definition.content_type and role_definition.content_type.model in ('organization', 'team'):
+                obj_resource = Resource.objects.get(ansible_id=assignment_tuple.ansible_id_or_pk)
+                role_definition.give_permission(actor, obj_resource.content_object)
+            else:
+                remote_obj = RemoteObject(role_definition.content_type, assignment_tuple.ansible_id_or_pk)
+                role_definition.give_permission(actor, remote_obj)
+            return True
+        except Resource.DoesNotExist:
```
john-westcott-iv, line 1665: "with ansible_id" is ambiguous -- could be the actor's or the content object's. Clarify which one:

```suggestion
            self.stderr.write(f"Warning: Unable to find content object with ansible_id {assignment_tuple.ansible_id_or_pk} for role {assignment_tuple.role_definition_name}, skipping assignment")
```

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+            self.stderr.write(f"Warning: Unable to find object with ansible_id {assignment_tuple.ansible_id_or_pk}, skipping assignment")
+            return False
         except Exception as e:
```
john-westcott-iv, line 1674: At scale, operators need to know *which* assignment failed. Include the tuple fields:

```suggestion
            self.stderr.write(f"Warning: Unable to give permission for {assignment_tuple.assignment_type} assignment (actor={assignment_tuple.actor_ansible_id}, role={assignment_tuple.role_definition_name}, object={assignment_tuple.ansible_id_or_pk}), skipping: {e}")
```

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+        service (non-zero exit for the installer to retry).
+
+        give_permission is idempotent (uses get_or_create), so replaying
+        a partial page after a crash is safe.
+        """
+        self.stdout.write(f"Migrating role assignments from {service_slug} (type {service_type_name})")
+        roles_to_exclude = self._get_role_definitions_to_exclude(service_type_name)
+
+        total_created = 0
+        for assignment_type in ('user', 'team'):
```
john-westcott-iv, line 1731: The scattered `if assignment_type == 'user' else` pattern is fragile -- if a third type is ever added, you'd have to find and update every occurrence. A dispatch dict in one place:

```python
LIST_METHODS = {
    'user': 'list_user_assignments',
    'team': 'list_team_assignments',
}

for assignment_type, method_name in LIST_METHODS.items():
    list_fn = getattr(self.client, method_name)
    cursor = _CursorStore(service_slug, assignment_type)
    ...
```

One place to add a new type. The old `AssignmentActorType` enum was heavier than needed, but a dispatch dict gives the same extensibility without the ceremony.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -325,11 +308,8 @@ def _report_migration_summary(
...
 
         self.stdout.write("\n=== Re-enabling service authentication ===")
-        if not self._services_with_count_drift:
-            MigrateServiceDataHasRan.mark_migration_completed()
-            self.stdout.write("✓ Migration flag updated: Service authentication is now enabled.")
-        else:
-            self.stdout.write("⚠ Migration flag NOT set due to count drift. Re-run to complete migration.")
+        MigrateServiceDataHasRan.mark_migration_completed()
+        self.stdout.write("✓ Migration flag updated: Service authentication is now enabled.")
         self.stdout.write("\nAll services migration completed successfully!")
```
john-westcott-iv: **Needs human judgment.** This is now set unconditionally. The old code withheld the flag when count drift was detected:

```python
if not self._services_with_count_drift:
    MigrateServiceDataHasRan.mark_migration_completed()
else:
    self.stdout.write("Migration flag NOT set due to count drift.")
```

With the PK cursor this may be intentional (the cursor tracks progress, so subsequent runs pick up from where they left off). But it changes the contract with the installer -- previously, drift meant non-zero exit + flag NOT set. Please confirm this behavioral change is acceptable to the installer team.

aap_gateway_api/models/migrate_data.py
```diff
@@ -3,6 +3,48 @@
 from django.db import models
 
 
+class MigrateServiceDataLastRolePK(models.Model):
```
john-westcott-iv: **Drop this model, the migration (0024), and the `__init__.py` import change.** The cursor is only used inside `migrate_service_data` -- no admin UI, no querysets, no foreign keys. A full ORM model with migration infrastructure is over-engineered for a key-value lookup of one integer.

Replace with a self-contained `_CursorStore` class inside the management command file:

```python
class _CursorStore:
    TABLE = 'migrate_service_data_role_cursor'

    def __init__(self, service_slug, assignment_type):
        self.service_slug = service_slug
        self.assignment_type = assignment_type
        self._ensure_table()
        self.last_pk = self._load()

    def _ensure_table(self):
        try:
            with connection.cursor() as c:
                c.execute(f"""
                    CREATE TABLE IF NOT EXISTS {self.TABLE} (
                        service_slug VARCHAR(255) NOT NULL,
                        assignment_type VARCHAR(4) NOT NULL,
                        last_pk BIGINT NOT NULL DEFAULT 0,
                        PRIMARY KEY (service_slug, assignment_type)
                    )
                """)
        except Exception:
            logger.warning("Could not create cursor table, will process all assignments")

    def _load(self):
        try:
            with connection.cursor() as c:
                c.execute(
                    f"SELECT last_pk FROM {self.TABLE} "
                    f"WHERE service_slug = %s AND assignment_type = %s",
                    [self.service_slug, self.assignment_type],
                )
                row = c.fetchone()
                return row[0] if row else 0
        except Exception:
            return 0

    def advance(self, pk):
        try:
            with connection.cursor() as c:
                c.execute(
                    f"""INSERT INTO {self.TABLE}
                        (service_slug, assignment_type, last_pk)
                        VALUES (%s, %s, %s)
                        ON CONFLICT (service_slug, assignment_type)
                        DO UPDATE SET last_pk = EXCLUDED.last_pk""",
                    [self.service_slug, self.assignment_type, pk],
                )
        except Exception:
            logger.warning("Could not save cursor position, next run will reprocess")
```

**Benefits:**
- **No migration on any branch** -- same code cherry-picks to 2.6, 2.7, devel without migration number conflicts
- **Self-contained** -- cursor persistence lives entirely in the command file
- **Graceful degradation** -- if any DB operation fails, falls back to processing all assignments (`give_permission` is idempotent, so the worst case is "slow")
- **`self.last_pk` is immutable after `__init__`** -- `advance()` only persists to DB, preventing the pagination bug

Also drop: migration `0024_migrateservicedatalastrolepk.py`, the import in `models/__init__.py`, and the `MigrateServiceDataLastRolePK` import in the command file.

john-westcott-iv (changes requested):
## Code Review: [AAP-78674] Replace role assignment migration with set-diff approach

**Score: 8.7/10 -- NEEDS_CHANGES**

| Lens | Score |
|---|---|
| Functionality | 7/10 (1 Critical) |
| Security | 10/10 |
| Quality | 9/10 |

### PR description vs implementation

The PR title and description describe a "set-diff approach" (build local/remote sets, compute delta). The actual implementation is a **PK-based cursor** (`id__gt=<last_pk>&order_by=id`). These are architecturally different strategies. The commit messages correctly describe the PK cursor. Please update the description to match the implementation -- the mismatch has already caused review confusion (see below).

### Note on the existing CHANGES_REQUESTED review

The "Major: Post-creation drift detection gap" finding from @john-westcott-iv appears to be based on the PR description rather than the actual code. The review states "ALL pages are fetched into a set before ANY creation happens" -- this is incorrect; `_paginate_and_create` creates assignments per page. The suggested fix ("page-at-a-time set-diff") describes what the code already does. The "drift gap" concern (items added during creation silently missed) doesn't apply to the PK cursor -- new items have higher PKs and are picked up on the next run.

### What's good

- The PK cursor architecture is sound -- 100-500x performance win for reinstalls
- HTTP status code checking before `.json()` is an improvement over the old code
- Test coverage is thorough (35 new tests)
- Clean deletion of old code with no stale references
- The E2E reinstall no-op test directly validates the core performance scenario

### Key changes needed

See inline comments for details:
1. **Critical:** Fix the pagination bug (page counter + cursor filter drift = items skipped)
2. Replace `MigrateServiceDataLastRolePK` Django model with a self-contained `_CursorStore` class -- eliminates migration files, enables clean backport to 2.6/2.7
3. Several simplifications: drop retry logic, drop intermediate methods, use dispatch dict for user/team, improve error messages

Full review output available on request.

---
*Generated with [Claude Code](https://claude.com/claude-code) / Opus 4.6 (Anthropic)*

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
-                try:
-                    if gateway_content_object:
-                        role_assignment = gateway_role_definition.give_permission(gateway_actor, gateway_content_object)
-                    else:
-                        role_assignment = gateway_role_definition.give_global_permission(gateway_actor)
-                    self.stdout.write(f"Gave permission: {self._format_migrated_assignment_for_logging(role_assignment)}")  # type: ignore
-                except Exception as e:
-                    self.stderr.write(f"Error: Unable to give permission for role {assignment_actor.value} assignment, skipping: {str(e)}")
-                    continue
+        try:
```
jay-steurer, line 1624: **[Minor | LOW confidence]** Edge case: if `role_definition.content_type` is `None` (global role) but `ansible_id_or_pk` is non-null (e.g., remote API returns `object_id` for a global assignment), this calls `RemoteObject(None, pk)` which will likely raise.

The exception IS caught by the generic `except Exception` below, so it fails with a warning rather than silently. No code change needed — this requires malformed remote data (a global role returning object references), which indicates a bug in the upstream service.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+                cursor.advance(last_pk_on_page)
+
+            if not data.get('next'):
                 break
             page += 1
 
-    def migrate_role_assignments(self, assignment_actor: AssignmentActorType, service_slug: str, service_type_name: str) -> None:
+        return created
+
+    def _create_page_assignments(self, results: List[Dict[str, Any]], assignment_type: str) -> int:
```
jay-steurer, line 1568: **[Nit]** Missing `service_slug` in drift warning — the old `_fetch_role_assignments` included it. When debugging count drift across multiple services, this log message does not identify which service experienced the change.

Suggestion: pass `service_slug` into `_paginate_assignments` and include it in the warning message.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1317,189 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+            filters = self._build_page_filters(page, roles_to_exclude, cursor)
+            response = self._fetch_page_with_retry(list_fn, assignment_type, filters)
+
+            data = response.json()
+            results = data.get('results') or []
+            if not results:
+                break
+
+            created += self._create_page_assignments(results, assignment_type)
+
```
jay-steurer, line 1550: **[Nit]** No try/except for network errors from `list_fn()`. The old `migrate_role_assignments` had a top-level `except Exception` that caught fetch failures with a descriptive "Unable to fetch role assignments" message. The new code relies on the caller's exception handling, producing less specific error output.

Not a defect — the exception propagates and is caught at a higher level. If desired, wrap the `list_fn` call to produce a more descriptive error message specific to the assignment type and page number.

jay-steurer (commented):
## Code Review: [AAP-78674] Replace role assignment migration with set-diff approach

**Score**: 9.8/10 (Functionality 9.5, Security 10, Quality 10)

No critical or major findings. See inline comments for 1 Minor (LOW confidence edge case) and 2 Nits (logging improvements).

### Needs Human Judgment
- Verify `get_local_assignments()` key resolution matches `_build_assignment_tuple` (DAB PR #1024 dependency)
- Confirm `_migrate_single_service` catches `RuntimeError` from persistent double-drift

### Observations
- Clean architecture improvement: O(N) give_permission calls → O(delta) set-diff
- Solid test coverage with 35 new unit tests + E2E reinstall test
- Security: all data paths use parameterized ORM queries. No injection surface.
- Two-pass drift recovery is well-designed with correct exit code semantics

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Intentional. No drift concept with cursors. Flag set when no services failed; CommandError raised before flag on any failure.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Fixed. Replaced with _CursorStore where self.last_pk is immutable. Base filters built once before the loop. Also dropped retry logic per your suggestion.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Exceptions propagate to _process_all_services top-level handler which includes the service slug. Per-method wrapper would duplicate without adding info.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: No longer applicable — drift tracking removed with PK cursor.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Theoretical edge case. Caught by except Exception with full identifiers in message. No change needed.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Fixed — now says "content object with ansible_id" with actor/role context.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Fixed — all error messages include actor, role, and object identifiers.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Kept inline ternary. Two types, no plans for a third, one readable line.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Agreed — removed. Merged into _create_assignment(dict, type). Removes the only role_sync_utils import.

aap_gateway_api/models/migrate_data.py
tznamena: Done. Raw SQL _CursorStore with CREATE TABLE IF NOT EXISTS and ON CONFLICT. Migration file deleted. Graceful degradation to last_pk=0.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1411,217 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+        total_created = 0
+        for assignment_type in ('user', 'team'):
+            list_fn = self.client.list_user_assignments if assignment_type == 'user' else self.client.list_team_assignments
+            cursor = _CursorStore(service_slug, assignment_type)
+            created = self._paginate_and_create(list_fn, assignment_type, roles_to_exclude, cursor)
+            total_created += created
+            self.stdout.write(f"  {assignment_type}: {created} assignments created")
+
+        self.stdout.write(f"Role assignment migration for {service_slug} completed ({total_created} total created)")
```
john-westcott-iv: **Drift detection needed for installer retry contract.**

The PK cursor eliminates the old `_services_with_count_drift` tracking, but the installer still relies on exit code to decide whether to retry. With the old code: count changed during pagination → drift detected → migration flag NOT set → non-zero exit → installer retries.

With the PK cursor: items added during the run get PKs higher than what's been processed. The cursor advances to the last item it saw, the flag is set, and the command exits 0. Those items *would* be caught on a re-run — but the installer won't trigger one because it got exit 0.

We have a customer upgrading at scale (~220K assignments) who is constantly hitting concurrent modifications. The installer needs to know the migration wasn't fully clean.

Suggested fix — a lightweight post-run check after each assignment type completes:

```python
        has_drift = False
        for assignment_type in ('user', 'team'):
            list_fn = ...
            cursor = _CursorStore(service_slug, assignment_type)
            created = self._paginate_and_create(list_fn, assignment_type, roles_to_exclude, cursor)
            total_created += created
            self.stdout.write(f"  {assignment_type}: {created} assignments created")

            # Post-run drift check: items added during pagination have higher PKs
            db_cursor = _CursorStore(service_slug, assignment_type)
            if db_cursor.last_pk > 0:
                check_resp = list_fn(filters={'order_by': 'id', 'id__gt': str(db_cursor.last_pk), 'page_size': '1'})
                if check_resp.status_code == 200 and check_resp.json().get('count', 0) > 0:
                    has_drift = True
                    self.stderr.write(
                        f"Warning: new {assignment_type} assignments appeared "
                        f"during migration of {service_slug} (concurrent modification)")

        ...
        if has_drift:
            raise CommandError(
                f"Role assignment migration for {service_slug} completed ({total_created} created) "
                f"but concurrent modifications were detected. Re-run to process remaining assignments.")
```

One extra API call per type — negligible overhead — but restores the installer's ability to detect incomplete state.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1411,217 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+            results = data.get('results') or []
+            if not results:
+                break
+
+            created += self._create_page_assignments(results, assignment_type)
+
+            # Advance cursor in DB for crash recovery.  cursor.last_pk
+            # is NOT mutated — base_filters stays consistent.
+            last_pk_on_page = results[-1].get('id')
+            if last_pk_on_page is not None:
```
john-westcott-iv: This guard can never trigger in practice. The `if not results: break` above guarantees `results` is non-empty, and every item came from an API ordered by `id`. If `results[-1].get('id')` is `None`, the API returned a role assignment without a primary key — the entire cursor model is broken at that point.

Silently swallowing this (cursor doesn't advance, next run reprocesses everything) would manifest as "migration keeps reprocessing all assignments and nobody knows why."

Suggest replacing with an explicit failure:

```suggestion
            last_pk_on_page = results[-1].get('id')
            if last_pk_on_page is None:
                raise RuntimeError(
                    f"API returned {assignment_type} assignment without 'id' field — "
                    f"cannot advance cursor"
                )
            cursor.advance(last_pk_on_page)
```

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1411,217 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+                    response.status_code,
+                    assignment_type,
+                    page,
                 )
-                total_count = new_count
-                self._services_with_count_drift.add(service_slug)
-            for assignment in json_response.get('results', []):
-                yield assignment
-            if not json_response.get('next'):
+                raise RuntimeError(f"Failed to fetch {assignment_type} assignments page {page}: HTTP {response.status_code}")
```
john-westcott-iv: At scale, a bare `HTTP 500` makes debugging difficult. The response body often contains the actual error (Django traceback, DRF error detail). Suggest capturing it:

```suggestion
                body_preview = ""
                try:
                    body_preview = response.text[:500]
                except Exception:
                    pass
                raise RuntimeError(
                    f"Failed to fetch {assignment_type} assignments page {page}: "
                    f"HTTP {response.status_code}\n{body_preview}"
                )
```

aap_gateway_api/tests/management/test_migrate_service_data_role_assignments.py
john-westcott-iv, line 1: These tests should live in the existing `test_migrate_resources.py`. That file is also misnamed (should be `test_migrate_service_data.py`), but having tests for the same management command split across two files makes them harder to find. Please consolidate into the existing file.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1409,98 +1411,217 @@ def _get_role_definitions_to_exclude(service_type: str) -> List[str]:
...
+                cursor.advance(last_pk_on_page)
+
+            if not data.get('next'):
                 break
             page += 1
 
-    def migrate_role_assignments(self, assignment_actor: AssignmentActorType, service_slug: str, service_type_name: str) -> None:
+        return created
+
+    def _create_page_assignments(self, results: List[Dict[str, Any]], assignment_type: str) -> int:
```
john-westcott-iv, line 1568: Not a blocker for this PR, but worth a follow-up optimization: each `_create_assignment` call does 3-5 individual DB queries (resolve role def, resolve actor, resolve content object, then `give_permission` with `get_or_create`). At 200 items per page, that's ~1000 queries per page.

A bulk approach could:
1. **Batch-fetch resolution** — `RoleDefinition.objects.filter(name__in=[...])`, `Resource.objects.filter(ansible_id__in=[...])` for all items on the page. This alone cuts queries from ~1000 to ~400.
2. **Bulk-create assignments** — DAB's own migration utility (`ansible_base/rbac/migrations/_utils.py`) uses `RoleUserAssignment.objects.bulk_create(assignments, ignore_conflicts=True)` with manually populated `object_id`, `content_type_id`, and `role_definition_id` fields. This could take it down to ~10-15 queries per page.

Caveats for full bulk_create: must manually populate cached fields from `AssignmentBase.save()`, must call `compute_object_role_permissions()` afterward to rebuild the RoleEvaluation cache, and audit logging signals are bypassed (acceptable for migration). Separate PR territory.

john-westcott-iv (commented):
The PK cursor approach is architecturally sound and addresses all prior review findings. Five inline comments cover:

1. **Drift detection** — the installer needs a non-zero exit when concurrent modifications are detected, otherwise it will not retry and items are permanently missed
2. **Dead guard on last_pk_on_page** — replace silent swallow with explicit RuntimeError
3. **HTTP response body** — include response.text in error for debugging at scale
4. **Test file consolidation** — merge into existing test_migrate_resources.py
5. **Bulk insert opportunity** — follow-up optimization (not a blocker)

Also: the PR title and description say "set-diff approach" but the implementation is a PK-based cursor (id__gt=<last_pk>&order_by=id). The commit messages correctly describe the PK cursor. This mismatch has already caused review confusion — my first review round had a Major finding based on the description that did not apply to the actual code. Please update the title/description to match the implementation before merge.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Agreed — the installer contract requires a signal for incomplete state. Implemented the post-run drift check as suggested: one extra API call per assignment type after pagination completes, querying id__gt<cursor>&page_size=1. If count > 0, a warning is logged and `RuntimeError` is raised so the installer retries.

Note: this captures items created between "last page fetched" and "this check." A tiny window remains between the check and the migration flag being set, but that gap is milliseconds and practically negligible.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Agreed — silently swallowing this would cause a cursor that never advances, manifesting as "migration keeps reprocessing all assignments and nobody knows why." Replaced the if guard with an explicit RuntimeError as suggested.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Done — extracted into _raise_fetch_error which captures up to 500 characters of response.text. Extracted as a separate method to keep _paginate_and_create under the cognitive complexity threshold of 15.

aap_gateway_api/tests/management/test_migrate_service_data_role_assignments.py
tznamena: The combined file would be ~3100 lines (2325 + 785). We think the current split has a clear boundary — unit tests for individual role assignment methods vs. end-to-end command tests. Single 3100-line file would be harder to navigate. 
I'm open to renaming the files for discoverability (e.g. both prefixed with test_migrate_service_data_) if that helps.

aap_gateway_api/management/commands/migrate_service_data.py
tznamena: Good idea — batch-fetch resolution and bulk_create with ignore_conflicts=True could cut queries from ~1000 to ~10-15 per page. Agreed this is follow-up territory — tracking it for a separate PR.

tznamena:
Note on PK cursor and the id field: 
The cursor optimization (resume from last-processed PK, instant completion on reinstall) depends on the service-index assignment API including `id` in responses. DAB's `assignment_common_fields` in `ansible_base/rbac/service_api/serializers.py` does not currently include id, so the cursor cannot advance. The migration works correctly without it, `give_permission` is idempotent, so reprocessing is safe. But the performance benefit requires a follow-up DAB change to add id to the service assignment serializers.

john-westcott-iv (changes requested):
## Re-Review: [AAP-78674] PK Cursor Role Assignment Migration

**Verdict: NEEDS_CHANGES** (one prerequisite blocker)

All findings from the previous two review rounds have been addressed — the `_CursorStore` design is solid, `last_pk` immutability is correct, error messages include full identifiers, retry logic is removed, and the architecture is sound.

### Blocker: DAB serializer prerequisite

The PK cursor optimization depends on the `id` field being present in assignment API responses. However, `migrate_service_data` calls the **service-index** API (`/api/v2/service-index/role-user-assignments/`), not the public API (`/api/v2/role_user_assignments/`). The service-index serializer in DAB (`assignment_common_fields` in `ansible_base/rbac/service_api/serializers.py`) does not include `id`.

This means the cursor never advances — every run reprocesses all assignments. The code handles this gracefully (warning + idempotent replay), but the headline performance win (reinstall in <1 second vs 4+ hours) does not materialize. Without the DAB change, this PR adds ~400 lines of cursor infrastructure that operates as a no-op.

**Required**: Add `'id'` to `assignment_common_fields` in `ansible_base/rbac/service_api/serializers.py` (django-ansible-base) and merge that change before this PR. Once the service-index API returns `id`, the cursor works as designed.

### Previously addressed (confirmed)

- Critical pagination bug (page counter + cursor drift) — fixed, `last_pk` immutable
- Django model replaced with raw-SQL `_CursorStore` — no migration needed
- `AssignmentTuple` removed, works with raw API dicts
- Per-page retry logic removed
- Post-run drift detection implemented
- HTTP error body preview in RuntimeError
- Error messages include actor/role/object identifiers
- Team assignment assertion in E2E reinstall test

---
*Generated with [Claude Code](https://claude.com/claude-code) / Opus 4.6 (Anthropic)*

tznamena:
The failure in aap-dev is unrelated to changes in this PR. Even running aap-dev locally with dab and gw local sources fails and the container running migrations fails with:
```
Failed library import, colors won't be in logs: No module named 'colorama'                                                                                                                   
             ^^^^^^^^                                                                                                                                                                        
  File "/app/gateway/aap_gateway_api/__init__.py", line 23, in manage                                                                                                                        
    execute_from_command_line(sys.argv)                                                                                                                                                      
  File "/opt/aap-gateway/venv/lib64/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line                                                     
    utility.execute()                                                                                                                                                                        
  File "/opt/aap-gateway/venv/lib64/python3.12/site-packages/django/core/management/__init__.py", line 382, in execute                                                                       
    settings.INSTALLED_APPS                                                                                                                                                                  
  File "/opt/aap-gateway/venv/lib64/python3.12/site-packages/django/conf/__init__.py", line 81, in __getattr__                                                                               
    self._setup(name)                                                                                                                                                                        
  File "/opt/aap-gateway/venv/lib64/python3.12/site-packages/django/conf/__init__.py", line 68, in _setup                                                                                    
    self._wrapped = Settings(settings_module)                                                                                                                                                
                    ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                
  File "/opt/aap-gateway/venv/lib64/python3.12/site-packages/django/conf/__init__.py", line 166, in __init__                                                                                 
    mod = importlib.import_module(self.SETTINGS_MODULE)                                                                                                                                      
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                      
  File "/usr/lib64/python3.12/importlib/__init__.py", line 90, in import_module                                                                                                              
    return _bootstrap._gcd_import(name[level:], package, level)                                                                                                                              
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                              
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import                                                                                                                            
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load                                                                                                                         
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked                                                                                                                
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked                                                                                                                          
  File "<frozen importlib._bootstrap_external>", line 999, in exec_module                                                                                                                    
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed                                                                                                               
  File "/app/gateway/aap_gateway_api/settings.py", line 24, in <module>                                                                                                                      
    load_python_file_with_injected_context(settings_file_path, settings=DYNACONF)                                                                                                            
  File "/app/django-ansible-base/ansible_base/lib/dynamic_config/dynaconf_helpers.py", line 290, in load_python_file_with_injected_context                                                   
    settings._loaded_files.append(file_path)               
    ^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                   
  File "/opt/aap-gateway/venv/lib64/python3.12/site-packages/dynaconf/base.py", line 98, in __getattr__                                                                                      
    return getattr(self._wrapped, name)                                                                                                                                                      
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                      
  File "/opt/aap-gateway/venv/lib64/python3.12/site-packages/dynaconf/base.py", line 369, in __getattr__                                                                                     
    raise AttributeError(key)                                                                                                                                                                
AttributeError: _loaded_files
```

john-westcott-iv (approved):

PR #141, opened 2026-06-26
fix(AAP-80948): add metrics to migrate_service_data SERVICE_TYPE_ORDER

cshiels-ie:
## Summary

- Add \`METRICS = \"metrics\"\` to the \`DefaultServiceType\` enum in \`service_type.py\`
- Add \`DefaultServiceType.METRICS.value\` to \`SERVICE_TYPE_ORDER\` in \`migrate_service_data.py\`
- Add metrics entry to \`_default_service_types\` in \`_initialize_tasks.py\` for fresh installs
- Fix \`conftest.py\` \`port_prefixes\` and \`_service_index_paths\` for the new service type

## Problem

\`migrate_service_data\` hard-codes \`SERVICE_TYPE_ORDER = [controller, hub, eda]\` and filters \`ServiceAPIRoute\` objects to only those three types. As a result, the \`service_id\` seeding step:

\`\`\`python
service_api.service_cluster.service_id = self.upstream_service_id
service_api.service_cluster.save()
\`\`\`

never runs for the metrics \`ServiceCluster\`. Its \`service_id\` column remains \`NULL\` after install.

\`validate_service_token()\` does \`ServiceCluster.objects.get(service_id=<iss_from_jwt>)\` — this lookup always misses a NULL row, raising \`DoesNotExist\` → HTTP 401 to the metrics service on every \`jwt_claims\` request. This prevents the metrics service from resolving user permissions for non-admin users, causing Platform Auditors (system auditors) to receive HTTP 403 on all metrics endpoints.

Additionally, the \`metrics\` \`ServiceType\` DB row was never seeded, so \`ServiceType.objects.get(name=\"metrics\")\` fails on fresh installs.

There is no alternative self-registration path:
- \`ServiceMetadataView\` (DAB) is GET-only — no POST endpoint exists
- \`ServiceCluster.service_id\` is \`editable=False\` — REST API silently ignores PATCH writes
- The metrics service has no outbound gateway registration code (\`RESOURCE_SERVER_SYNC_ENABLED = False\`)

Tracked in: **AAP-80948**

## Fix

Adding \`METRICS\` to \`DefaultServiceType\` and \`SERVICE_TYPE_ORDER\` causes \`migrate_service_data\` to include the metrics service in its loop. The \`service_id\` is seeded from the metadata endpoint response. The org/team/user migration steps run but handle empty result sets cleanly.

The \`metrics\` \`ServiceType\` row is seeded via \`_default_service_types\` in \`_initialize_tasks.py\` so fresh installs get it through migration \`0007\`. The installer creates this row before \`migrate_service_data\` runs, so existing installs are also covered without a separate data migration.

## Test plan

- [ ] Run \`migrate_service_data\` against an install that includes metrics — verify \`aap_gateway_api_servicecluster.service_id\` is populated for the metrics row after the command completes
- [ ] Confirm a non-admin Platform Auditor user can authenticate to metrics endpoints without receiving a 403
- [ ] Confirm existing controller/hub/eda migration behaviour is unchanged

## AI Assistance
This change was developed with assistance from Claude Code (Claude Sonnet 4.6).
All generated code was reviewed, tested, and validated before merging.

## Summary by CodeRabbit

* **New Features**
  * Added support for the Metrics service type.
  * Metrics services are now initialized with standard health-check and service-index endpoints.
  * Service data migration now includes Metrics alongside existing platform services.

* **Bug Fixes**
  * Ensured Metrics service routes and local service configuration are handled consistently during setup.

* **Tests**
  * Added coverage verifying Metrics is included in the service migration sequence.


aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -167,6 +167,7 @@ class Command(BaseCommand):
         DefaultServiceType.CONTROLLER.value,
         DefaultServiceType.HUB.value,
         DefaultServiceType.EDA.value,
+        DefaultServiceType.METRICS.value,
```
Dostonbek1, line 76: Nit: the comment on line 165 (`# Service processing order - Controller first to establish priority for user merging`) is fine, but the inline comment at L396 (not shown in diff, but visible in full source) says:

```python
# Get all services with DefaultServiceType in exact order: controller, hub, eda
```

Now that metrics is included, this comment is stale. Consider updating it or removing the explicit list from the comment so it doesn't drift again:

```suggestion
        DefaultServiceType.METRICS.value,
    ]
```

(The suggestion itself is a no-op — this is just a reminder to update the comment on L396.)

Dostonbek1 (commented):
## Code Review Summary

Clean, well-targeted fix for AAP-80948. Adding `METRICS` to `DefaultServiceType` and `SERVICE_TYPE_ORDER` correctly ensures that `migrate_service_data` seeds the `service_id` on the metrics `ServiceCluster`, which is required for JWT validation. The initialization task entry for fresh installs and the conftest fixture updates are correct and complete.

**Files Reviewed**: 4 files
**Comments Posted**: 1 review comment

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 1 code quality suggestion (stale comment)

### Overall Assessment
LGTM — minor nit only.

### General Feedback
- Good defensive design: `_migrate_single_service` handles services that don't expose orgs/teams/users gracefully via `_is_service_already_synced()` and empty pagination results.
- The `login_path: None` / `logout_path: None` choice for metrics is correct since metrics has no user-facing auth flow.
- The `service_index_path: "/v1/service-index/"` aligns with EDA and the operator ConfigMap template, enabling RBAC push URLs to be constructed properly.
- The failing CI test (`test_org_admin_can_create_unprivileged_user`) appears to be a pre-existing flaky test unrelated to this change — it fails consistently across all runs on `devel`.

jeffh-oss:
@john-westcott-iv and @bhavenst we were discussing this in the weekly PR review meeting and we feel like one or both of you should have a look at this.

bhavenst (commented):
## Review Summary

Clean, well-targeted fix. The code change is correct — `_migrate_single_service` handles metrics gracefully (empty resource sets are skipped via `_is_service_already_synced()`, empty role assignment pagination breaks cleanly, superuser sync correctly ignores non-controller/hub/eda types). `login_path: None` / `logout_path: None` and the `service_index_path` value are all correct.

Two observations (not blocking):

**1. No test covers the core fix.** `test_service_processing_order` hardcodes `len(processing_lines) == 3` and only checks controller/hub/eda. If someone removed `METRICS` from `SERVICE_TYPE_ORDER`, no test would fail. A small assertion like `assert "metrics" in Command.SERVICE_TYPE_ORDER` would directly validate the fix.

**2. PR description mentions migration 0024 but it's not in the diff.** This isn't breaking — the installer creates the `ServiceType` row before calling `migrate_service_data`, so the row exists when needed and the `ServiceAPIRoute` filter on line 198 simply won't match metrics if the row is absent. But the description should either be updated or the migration added for defense-in-depth.

Minor: the `test_ping.py` parentheses removal is valid but unrelated to AAP-80948.

Verified Ciaran's claim from the Jira discussion: `migrate_service_data` (via `service_orchestration.py:90-91`) is indeed the **only** ORM write path for non-gateway `ServiceCluster.service_id`. The field is `editable=False` (API ignores PATCH writes), `ServiceMetadataView` is GET-only, the installer doesn't set it, and the `save()` override only auto-populates for the gateway type. The jewel PR is needed regardless of the installer fix (MR !1187) — without `service_id`, metrics' service JWT is rejected by `validate_service_token()` and `resource_sync` (PR #320) can't authenticate either.

bhavenst (approved):
There are a few minor improvement suggestions, but up to you if you want to add them or not.  Approving to make it mergeable either way.

PR #142, opened 2026-06-29
[AAP-80669] rbac service cursor pagination

AlexSCorey:
This is a counterpart to https://github.com/ansible/django-ansible-base/pull/1038 and https://github.com/ansible/django-ansible-base/pull/1025 (merged already)

PR 1038 should be merged first, and this is pr can do it.
Requires: https://github.com/ansible/django-ansible-base/pull/1038

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of role-assignment migration by following upstream cursor pagination and advancing cursor state correctly across pages.
  * Prevented premature termination and missed records when processing multi-page assignments.
  * Enhanced drift detection by using returned `results` data instead of relying on response count fields.
* **Tests**
  * Updated mocked API responses to reflect cursor-style pagination (`next`/`results`) and removed `count`-based assertions.
  * Added coverage ensuring pagination stops after `next` lacks cursor/page parameters.

### DIFF
--- a/aap_gateway_api/management/commands/migrate_service_data.py
+++ b/aap_gateway_api/management/commands/migrate_service_data.py
@@ -1,6 +1,7 @@
 import logging
 from collections import OrderedDict
 from typing import Any, Dict, List, Optional, Tuple, Type
+from urllib.parse import parse_qs, urlparse
 
 from ansible_base.authentication.models import AuthenticatorUser
 from ansible_base.authentication.models.authenticator import Authenticator
@@ -1525,10 +1526,11 @@ class Command(BaseCommand):
         Crash safety: the cursor is advanced in the database after each
         fully-processed page, so at most one page of work is lost.
         """
-        page = 1
+        page_num = 1
         created = 0
+        next_cursor = None
 
-        # Build base filters once — only 'page' changes between iterations.
+        # Build base filters once.
         # snapshot_pk is immutable (set once in _CursorStore.__init__).
         base_filters: Dict[str, Any] = {'order_by': 'id', **self.BIG_PAGE_FILTERS}
         if roles_to_exclude:
@@ -1537,9 +1539,12 @@ class Command(BaseCommand):
             base_filters['id__gt'] = str(cursor.last_pk)
 
         while True:
-            response = list_fn(filters={**base_filters, 'page': page})
+            filters = {**base_filters}
+            if next_cursor:
+                filters['cursor'] = next_cursor
+            response = list_fn(filters=filters)
             if response.status_code != 200:
-                self._raise_fetch_error(response, assignment_type, page)
+                self._raise_fetch_error(response, assignment_type, page_num)
 
             data = response.json()
             results = data.get('results') or []
@@ -1559,9 +1564,13 @@ class Command(BaseCommand):
                 )
             cursor.advance(last_pk_on_page)
 
-            if not data.get('next'):
+            next_url = data.get('next')
+            if not next_url:
                 break
-            page += 1
+            next_cursor = parse_qs(urlparse(next_url).query).get('cursor', [None])[0]
+            if not next_cursor:
+                break
+            page_num += 1
 
         return created
 
@@ -1769,7 +1778,7 @@ class Command(BaseCommand):
 
         try:
             check_resp = list_fn(filters={'order_by': 'id', 'id__gt': str(db_cursor.last_pk), 'page_size': '1'})
-            if check_resp.status_code == 200 and check_resp.json().get('count', 0) > 0:
+            if check_resp.status_code == 200 and len(check_resp.json().get('results', [])) > 0:
                 self._log(f"Warning: new {assignment_type} assignments appeared during migration of {service_slug} (concurrent modification)", logging.WARNING)
                 return True
         except Exception:
--- a/aap_gateway_api/management/commands/migrate_service_data.py
+++ b/aap_gateway_api/management/commands/migrate_service_data.py
@@ -43,10 +43,11 @@ class _CursorStore:
     trivial to backport across versions.
 
     Key invariant: self.last_pk is set once in __init__ and NEVER
-    mutated.  advance() only persists progress to the database.
-    This ensures the HTTP id__gt filter stays immutable across all
-    pages of a single run, preventing items from being skipped when
-    the cursor advances in the database between pages.
+    mutated.  advance() only persists progress to the database
+    and updates last_advanced_pk in memory.  This ensures the HTTP
+    id__gt filter stays immutable across all pages of a single run,
+    while last_advanced_pk tracks the actual high-water mark for
+    drift detection without requiring a DB re-read.
 
     If any database operation fails, the store degrades gracefully
     to last_pk=0, causing the command to reprocess all assignments.
@@ -62,6 +63,7 @@ class _CursorStore:
         self.service_slug = service_slug
         self.assignment_type = assignment_type
         self.last_pk = self._load()
+        self.last_advanced_pk = self.last_pk
 
     def _ensure_table(self):
         """Create the cursor table if it does not already exist."""
@@ -105,10 +107,13 @@ class _CursorStore:
 
         The in-memory last_pk stays at its initial value so the HTTP
         id__gt filter remains consistent across all pages of a run.
+        last_advanced_pk is updated unconditionally so drift detection
+        can use the actual high-water mark without a DB re-read.
         If the upsert fails, a warning is logged but the run continues —
         the next invocation will reprocess from the old position, which
         is safe because give_permission is idempotent.
         """
+        self.last_advanced_pk = pk
         try:
             with connection.cursor() as cur:
                 cur.execute(
@@ -1508,34 +1513,30 @@ class Command(BaseCommand):
         raise RuntimeError(f"Failed to fetch {assignment_type} assignments page {page}: HTTP {response.status_code}\n{body_preview}")
 
     def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude: List[str], cursor: '_CursorStore') -> int:
-        """Paginate one assignment endpoint using cursor pagination, creating assignments per page.
+        """Paginate one assignment endpoint, creating assignments per page.
 
-        Uses cursor-based pagination where results are ordered by ``id``
-        (ascending), so the last result on each page has the highest PK.
-        The cursor is advanced per page so the drift check knows where
-        this run ended and crash recovery can resume from the last
-        completed page.
+        Follows API pagination tokens from the ``next`` URL while tracking
+        progress via a PK watermark (``_CursorStore``) for crash recovery.
+        Results are ordered by ``id`` (ascending) by DAB's
+        ``ServiceCursorPagination``, so the last result on each page has
+        the highest PK.
 
-        On a reinstall where the cursor is already past the last PK,
+        On a reinstall where the watermark is already past the last PK,
         the first page returns zero results and the method returns
         immediately.
 
-        No per-page retry — the PK cursor provides crash recovery.
-        If the command fails on an HTTP error, the installer re-runs
-        it and the cursor resumes from the last completed page.
-
-        Crash safety: the cursor is advanced in the database after each
-        fully-processed page, so at most one page of work is lost.
-        Since give_permission is idempotent, replayed assignments are
-        harmless.
+        Crash safety: the watermark is advanced in the database after
+        each fully-processed page, so at most one page of work is
+        replayed on re-run. Since give_permission is idempotent,
+        replayed assignments are harmless.
         """
         page_num = 1
         created = 0
         pagination_token = None
         last_pk = None
 
-        # Build base filters once.
-        # snapshot_pk is immutable (set once in _CursorStore.__init__).
+        # Build base filters once.  cursor.last_pk is immutable (set once
+        # in _CursorStore.__init__), so id__gt stays consistent across pages.
         base_filters: Dict[str, Any] = self._get_base_filters(roles_to_exclude, cursor)
 
         while True:
@@ -1553,15 +1554,10 @@ class Command(BaseCommand):
 
             created += self._create_page_assignments(results, assignment_type)
 
-            # Advance cursor in DB for crash recovery.  cursor.last_pk
-            # is NOT mutated — base_filters stays consistent.
+            # Advance watermark in DB for crash recovery.
             last_pk = results[-1].get('id')
             if last_pk is None:
-                raise RuntimeError(
-                    f"API returned {assignment_type} assignment without 'id' field — "
-                    f"cannot advance cursor. Check that the upstream service "
-                    f"is running a compatible DAB version."
-                )
+                raise RuntimeError(f"API returned {assignment_type} assignment without 'id' field — cannot advance watermark.")
             cursor.advance(last_pk)
             pagination_token, is_last = self._is_last(data)
             if is_last:
@@ -1573,18 +1569,18 @@ class Command(BaseCommand):
     def _get_base_filters(self, roles_to_exclude: List[str], cursor: '_CursorStore') -> Dict[str, Any]:
         """Build base filters for assignment pagination.
 
-        This method is called once per assignment type to construct the
-        base filters used in the pagination loop.  It sets up the
-        'order_by', 'id__gt', and 'not__role_definition__name__in' filters
-        based on the cursor and roles to exclude.
+        Called once per assignment type.  Sets up 'id__gt' and
+        'not__role_definition__name__in' filters based on the watermark
+        and roles to exclude.
 
         Args:
             roles_to_exclude: List of role names to exclude from the results
             cursor: CursorStore instance tracking the last processed PK
         """
-
-        # Note: ordering is controlled by ServiceCursorPagination (ordering = 'id')
-        # in DAB, not by a query param.
+        # NOTE: order_by here is redundant — ServiceCursorPagination in DAB
+        # controls the actual ordering and ignores this query param. We
+        # include it for readability but it has no effect on cursor-paginated
+        # endpoints.
         base_filters: Dict[str, Any] = {**self.BIG_PAGE_FILTERS, 'order_by': 'id'}
         if roles_to_exclude:
             base_filters['not__role_definition__name__in'] = ','.join(roles_to_exclude)
@@ -1786,7 +1782,7 @@ class Command(BaseCommand):
             # fetched" and "this check."  A tiny window remains between
             # this check and the migration flag being set, but that gap
             # is milliseconds and practically negligible.
-            if self._check_for_drift(list_fn, assignment_type, service_slug):
+            if self._check_for_drift(list_fn, assignment_type, service_slug, cursor):
                 drift_detected = True
 
         self._log(f"Role assignment migration for {service_slug} completed ({total_created} total created)", logging.INFO)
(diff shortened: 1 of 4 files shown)


aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1548,23 +1552,58 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
-            if not data.get('next'):
+            if page_num == 1:
+                highest_pk = results[0].get('id')
+                if highest_pk is None:
+                    raise RuntimeError(
+                        f"API returned {assignment_type} assignment without 'id' field — "
+                        f"cannot advance cursor. Check that the upstream service "
+                        f"is running a compatible DAB version."
+                    )
+                cursor.advance(highest_pk)
```
john-westcott-iv: **Major — Crash recovery bug: descending ordering breaks incremental resume**

With `ServiceCursorPagination` using `ordering = '-id'` (descending), page 1 returns the **highest** PKs. `cursor.advance(highest_pk)` on page 1 writes the global maximum PK to the DB.

If the command crashes after page 2:
- Re-run loads `last_pk = 100` (the highest PK) from DB
- `id__gt=100` returns nothing
- Items on later pages (lower PKs) are silently lost

The old ascending approach worked because `results[-1]` was the highest PK *processed so far*, and `id__gt` on re-run picked up right after it.

The fix likely needs to happen upstream in DAB — `ServiceCursorPagination` needs ascending ordering (`ordering = 'id'`) for the service-index assignment endpoints, so the `_CursorStore` per-page advance + `id__gt` resume design works correctly.

Moving `cursor.advance()` after the loop would avoid the stale-watermark problem but at the cost of replaying *all* pages on every crash — with 100k+ assignments that is a lot of (safe but slow) idempotent work.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1548,23 +1552,58 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
+        This method is called once per assignment type to construct the
+        base filters used in the pagination loop.  It sets up the
+        'order_by', 'id__gt', and 'not__role_definition__name__in' filters
+        based on the cursor and roles to exclude.
+
+        Args:
+            roles_to_exclude: List of role names to exclude from the results
+            cursor: CursorStore instance tracking the last processed PK
+        """
+        base_filters: Dict[str, Any] = {'order_by': 'id', **self.BIG_PAGE_FILTERS}
```
john-westcott-iv: **Minor — Remove `order_by: 'id'` — CursorPagination silently ignores it**

DRF's `CursorPagination` enforces its own ordering (set via `ServiceCursorPagination.ordering = '-id'`) and ignores any ordering applied by filter backends. DAB's `OrderByBackend` does apply `order_by=id` to the queryset, but `CursorPagination.paginate_queryset()` overrides it with `-id`.

Leaving `order_by: 'id'` in `base_filters` is actively misleading — a reader would reasonably assume ascending order is being enforced. Remove it so nobody hits a "but I sent the order_by param?" moment.

Same applies to line 1810 in `_check_for_drift`.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1507,39 +1508,42 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
...
-        fully-processed page, so at most one page of work is lost.
+        Crash safety: the cursor is advanced in the database after the
+        first page is processed. If the command fails mid-run, the next
+        invocation replays from the beginning of this run. Since
+        give_permission is idempotent, replayed assignments are harmless.
         """
-        page = 1
+        page_num = 1
         created = 0
+        next_cursor = None
```
john-westcott-iv: **Nit — Two different things called "cursor" is confusing**

This file now has two unrelated cursor concepts:
- `next_cursor` (line 1533/1543) — the API pagination token extracted from the `?cursor=` URL param
- `cursor` (the `_CursorStore` parameter) — a DB-persisted PK watermark for crash recovery

Consider renaming one to make the distinction obvious. For example, `page_token` for the API pagination value, or `watermark`/`pk_watermark` for the DB-persisted one.

aap_gateway_api/tests/views/api/v1/ping/test_ping.py
```diff
@@ -112,7 +112,7 @@ def test_ping_db_check_flushes_stale_connections(mock_close, mock_conns, request
     call_order = []
     mock_close.side_effect = lambda: call_order.append("close")
     cursor_cm = mock.MagicMock()
-    cursor_cm.__enter__ = mock.Mock(side_effect=lambda: (call_order.append("cursor") or mock.Mock()))
+    cursor_cm.__enter__ = mock.Mock(side_effect=lambda: call_order.append("cursor") or mock.Mock())
```
john-westcott-iv, line 115: **Nit — Unrelated change, should be a separate commit**

This parentheses cleanup in `test_ping.py` is unrelated to cursor pagination. Mixing cosmetic fixes with a pagination overhaul makes the PR harder to review and bisect. Consider splitting this into its own commit.

aap_gateway_api/management/commands/migrate_service_data.py
john-westcott-iv: Concrete suggestion — remove `order_by` and add a comment explaining why it's absent:

```suggestion
        # Note: ordering is controlled by ServiceCursorPagination (ordering = '-id')
        # in DAB, not by a query param. DRF's CursorPagination silently ignores
        # any order_by sent as a filter — do not add one here.
        base_filters: Dict[str, Any] = {**self.BIG_PAGE_FILTERS}
```

Same for the drift check on line 1810 — drop the `'order_by': 'id'` there too and add a similar note.

john-westcott-iv (changes requested):
## Code Review: [AAP-80669] rbac service cursor pagination

**Verdict**: NEEDS_CHANGES

### Key Findings

1. **Major — Crash recovery is broken by descending ordering.** `ServiceCursorPagination` in DAB uses `ordering = '-id'` (descending), so page 1 returns the highest PKs. `cursor.advance(highest_pk)` on page 1 writes the global maximum to the DB. On crash + re-run, `id__gt=<max>` returns nothing — items on later pages (lower PKs) are silently lost. The `_CursorStore` per-page advance + `id__gt` resume design requires ascending ordering. Fix likely needs to happen in DAB.

2. **Minor — `order_by: 'id'` in base_filters is silently ignored.** DRF's `CursorPagination` enforces its own ordering and overrides anything applied by filter backends. Leaving `order_by: 'id'` in the query params is misleading — readers will assume ascending order is being enforced when it isn't. Remove it.

3. **Nit — Two unrelated concepts both called "cursor"** makes the code harder to follow for humans. Consider renaming one (e.g. `page_token` vs `watermark`).

4. **Nit — `test_ping.py` change is unrelated** to cursor pagination and should be a separate commit.

### Observations

- **Drift check fails open (pre-existing):** `_check_for_drift` catches all exceptions and returns False ("assume no drift"). Not introduced by this PR, but worth a follow-up since it undermines the mid-run addition detection.
- Test coverage for the new pagination logic is solid.
- The `_is_last` / `_get_base_filters` extraction is clean.

aap_gateway_api/tests/management/test_migrate_service_data_role_assignments.py
```diff
@@ -230,13 +229,13 @@ def test_cursor_advances_per_page(self):
         """
         page1 = _make_api_response(
             [
-                _make_remote_assignment("user", "u1", "Role1", pk=10),
+                _make_remote_assignment("user", "u1", "Role1", pk=20),
```
jay-steurer: **[Quality — Minor]** Test name `test_cursor_advances_per_page` is misleading after the behavioral change. The cursor now advances **once** (on page 1 only), not per-page. The test body was correctly updated to reflect the new behavior (pk=20 on page 1, pk=10 on page 2 — descending), but the name still implies per-page advancement.

```suggestion
    def test_cursor_advances_on_first_page(self):
```

(Note: the suggestion above applies to the method definition line ~229, shown here for context.)

jay-steurer (commented):
## Code Review: [AAP-80669] rbac service cursor pagination (#142)

**Score**: 9.8/10 (Functionality 10, Security 10, Quality 9.5)

No critical or major findings. One minor: test name `test_cursor_advances_per_page` should be renamed (see inline comment).

### Needs Human Judgment

1. **`order_by=id` in `_get_base_filters` — dead code?** Cursor pagination ignores this parameter (tests confirm `-id` ordering). If intentionally kept for forward compatibility, consider adding a comment. If dead code, consider removing.

2. **Crash recovery semantics changed.** The cursor is now advanced to the highest PK on page 1. A crash on page N>1 means the next run fetches `id__gt=highest_pk` — which returns nothing (all remaining items have lower PKs). The docstring says "replayed assignments are harmless" but they aren't replayed — they're skipped. Aligns with CodeRabbit's finding #1.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1769,7 +1812,7 @@ def _check_for_drift(self, list_fn, assignment_type: str, service_slug: str) ->
 
         try:
```
john-westcott-iv, line 1807: **Still open from previous review** — `order_by: 'id'` was removed from `_get_base_filters` (good), but it's still here in `_check_for_drift`. Same issue applies: `CursorPagination` silently ignores it, so this is misleading. Drop it and add the same comment as in `_get_base_filters`.

```suggestion
            check_resp = list_fn(filters={'id__gt': str(db_cursor.last_pk), 'page_size': '1'})
```

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1548,23 +1552,62 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
+    def _get_base_filters(self, roles_to_exclude: List[str], cursor: '_CursorStore') -> Dict[str, Any]:
+        """Build base filters for assignment pagination.
+
+        This method is called once per assignment type to construct the
+        base filters used in the pagination loop.  It sets up the
+        'order_by', 'id__gt', and 'not__role_definition__name__in' filters
+        based on the cursor and roles to exclude.
+
+        Args:
+            roles_to_exclude: List of role names to exclude from the results
```
john-westcott-iv, line 1577: **Stale docstring** — this still says:

> It sets up the 'order_by', 'id__gt', and 'not__role_definition__name__in' filters

But `order_by` was just removed from this method. Update to match:

```suggestion
        'id__gt', and 'not__role_definition__name__in' filters
```

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1548,23 +1552,62 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
+            cursor: CursorStore instance tracking the last processed PK
+        """
+
+        # Note: ordering is controlled by ServiceCursorPagination (ordering = '-id')
+        # in DAB, not by a query param.
+        base_filters: Dict[str, Any] = {**self.BIG_PAGE_FILTERS}
+        if roles_to_exclude:
+            base_filters['not__role_definition__name__in'] = ','.join(roles_to_exclude)
+        if cursor.last_pk > 0:
```
john-westcott-iv, line 1587: **Ruff failure** — there are two blank lines between the docstring closing and the comment. This is likely what's failing `api-ruff-check` and `api-ruff-format`. Should be one blank line (or none).

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1548,23 +1552,62 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
+                    raise RuntimeError(
+                        f"API returned {assignment_type} assignment without 'id' field — "
+                        f"cannot advance cursor. Check that the upstream service "
+                        f"is running a compatible DAB version."
+                    )
+            pagination_token, is_last = self._is_last(data)
+            if is_last:
                 break
-            page += 1
+            page_num += 1
```
john-westcott-iv, line 1565: **Crash recovery — data loss fixed, but incremental resume lost**

Moving `cursor.advance()` after the loop fixes the silent data loss bug from the previous review — good. But this means every crash replays *all* pages from the previous watermark instead of resuming from the last completed page.

With the old ascending-order approach, `cursor.advance()` was called per page, so at most one page of work was replayed on crash. With descending order from `ServiceCursorPagination`, per-page advance doesn't work (the first page has the highest PKs, so advancing there skips everything below).

This trade-off (safe but slow replays) is acceptable as a stopgap, but the real fix is ascending ordering in DAB's `ServiceCursorPagination` so per-page advance works again. Is there a follow-up ticket for that?

john-westcott-iv (changes requested):
## Follow-up review

Thanks for the updates — naming fix (pagination_token) and test_ping.py removal look good. A few items still need attention:

### Still open
1. **`order_by: 'id'` still in `_check_for_drift`** (line 1813) — was removed from `_get_base_filters` but missed here
2. **Stale docstring** in `_get_base_filters` — still references `order_by` which was removed
3. **CI failing** — `api-ruff-check` and `api-ruff-format` are red, likely the extra blank line in `_get_base_filters`

### Discussion
4. **Crash recovery trade-off** — moving `cursor.advance()` after the loop fixes the data loss bug, but loses incremental resume (every crash replays all pages). Acceptable as a stopgap, but the root fix is ascending ordering in DAB's `ServiceCursorPagination`. Is there a follow-up ticket?

aap_gateway_api/management/commands/migrate_service_data.py
john-westcott-iv: Concrete example of why descending order breaks incremental resume:

**Old approach (ascending, per-page advance):**
```
Run 1 — 100 assignments exist, ascending order:
  Page 1: ids 1-20   -> cursor.advance(20)
  Page 2: ids 21-40  -> cursor.advance(40)
  ** CRASH **
  Re-run: id__gt=40  -> picks up at id 41
  Result: at most 1 page of work replayed
```

**New approach (descending, advance after loop):**
```
Run 1 — 100 assignments exist, descending order:
  Page 1: ids 100-81  (highest_pk = 100, not yet advanced)
  Page 2: ids 80-61
  ** CRASH **
  Re-run: cursor still at old value (0) -> replays ALL 100 assignments
  Result: safe (idempotent) but 5x slower than needed
```

**Why per-page advance can't work with descending:**
```
Hypothetical — descending order WITH per-page advance:
  Page 1: ids 100-81  -> cursor.advance(100)
  Page 2: ids 80-61
  ** CRASH **
  Re-run: id__gt=100 -> returns NOTHING
  Result: ids 60-1 silently lost (the bug in the previous revision)
```

Ascending order + per-page advance is the only combination that gives both safety and efficiency.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1507,12 +1508,13 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
         raise RuntimeError(f"Failed to fetch {assignment_type} assignments page {page}: HTTP {response.status_code}\n{body_preview}")
 
     def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude: List[str], cursor: '_CursorStore') -> int:
-        """Paginate one assignment endpoint using a PK cursor, creating assignments per page.
+        """Paginate one assignment endpoint using cursor pagination, creating assignments per page.
```
john-westcott-iv: **Nit — docstring conflates two cursor concepts**

The docstring says "cursor pagination" (the API pagination token) but the PK cursor (`_CursorStore`) is still doing the heavy lifting for crash recovery and drift detection. Lines 1514/1515 then say "The cursor is advanced per page" — which cursor? The reader has to guess it means the `_CursorStore`, not the API token.

Suggestion: acknowledge both mechanisms upfront, e.g.:

> Paginate one assignment endpoint using API cursor tokens, tracking progress via a PK watermark (`_CursorStore`) for crash recovery.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1524,22 +1526,25 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
         """
-        page = 1
+        page_num = 1
         created = 0
+        pagination_token = None
+        last_pk = None
 
-        # Build base filters once — only 'page' changes between iterations.
+        # Build base filters once.
         # snapshot_pk is immutable (set once in _CursorStore.__init__).
```
john-westcott-iv: **Nit — vestigial comment**

`snapshot_pk` is not a variable name in this code — it's left over from a previous revision. The actual variable is `cursor.last_pk`. Either update the reference or remove the comment since `_get_base_filters` already documents this contract.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1550,21 +1555,58 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
+        based on the cursor and roles to exclude.
+
+        Args:
+            roles_to_exclude: List of role names to exclude from the results
+            cursor: CursorStore instance tracking the last processed PK
+        """
+
+        # Note: ordering is controlled by ServiceCursorPagination (ordering = 'id')
+        # in DAB, not by a query param.
+        base_filters: Dict[str, Any] = {**self.BIG_PAGE_FILTERS, 'order_by': 'id'}
```
john-westcott-iv, line 1584: **Minor — `order_by` is silently ignored by CursorPagination**

DRF's `CursorPagination` enforces its own ordering (now `ordering = 'id'` in DAB's `ServiceCursorPagination`) and overrides anything applied by filter backends. This `order_by: 'id'` query param is accepted by DAB's `OrderByBackend` but then immediately overridden by `CursorPagination.paginate_queryset()`.

It happens to be consistent now (both ascending), so it's not wrong — just misleading. Add a comment so nobody assumes this is what controls the ordering:

```suggestion
        # NOTE: order_by here is redundant — ServiceCursorPagination in DAB
        # controls the actual ordering and ignores this query param. We include
        # it for readability but it has no effect on cursor-paginated endpoints.
        base_filters: Dict[str, Any] = {**self.BIG_PAGE_FILTERS, 'order_by': 'id'}
```

Same applies to line 1813 in `_check_for_drift`.

aap_gateway_api/tests/views/api/v1/ping/test_ping.py
john-westcott-iv, line 115: **Nit — unrelated change, was removed last revision but is back**

This parentheses cleanup is not related to cursor pagination. It was removed from the PR in the previous revision but has returned. Consider splitting into its own commit for clean bisect history.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1769,7 +1811,7 @@ def _check_for_drift(self, list_fn, assignment_type: str, service_slug: str) ->
```
john-westcott-iv, line 1806: **Two concerns with drift detection (pre-existing, flagging for follow-up):**

**1. Silent failure when cursor fails to load:**

`_CursorStore._load()` returns 0 for both "fresh install, no row" and "DB error, exception swallowed." Line 1806 treats both as "nothing to check":

```python
if db_cursor.last_pk <= 0:
    return False
```

If `advance()` wrote pk=40 to the DB but then `_load()` fails here (DB hiccup), drift detection is silently skipped and migration completes unvalidated. Consider distinguishing "no row" from "load error" — e.g. `_load()` could raise on exception instead of returning 0.

**2. Race window between last page and drift check:**

Between fetching/processing the last page (up to 200 items with `give_permission` calls) and running this drift check, new assignments could appear. With auto-increment PKs those get IDs above the watermark, so `id__gt` catches them. But consider re-loading the last page one final time as a belt-and-suspenders check — it would confirm nothing changed in the window between fetch and completion.

Neither is introduced by this PR — flagging for a follow-up ticket.

john-westcott-iv (commented):
## Review pass 3

The critical crash recovery bug is fixed — per-page advance with ascending ordering is correct. CI is passing. Nice work getting the DAB ordering change in.

### Actionable for this PR
1. **Docstring at line 1511** — conflates API cursor and PK watermark; clarify both mechanisms
2. **Line 1538** — vestigial `snapshot_pk` comment (variable is `cursor.last_pk`)
3. **Line 1588** — add a comment noting `order_by` is redundant because `CursorPagination` ignores it (same for line 1813 in `_check_for_drift`)
4. **test_ping.py** — unrelated drive-by fix is back; separate commit preferred

### Follow-up items (not blocking this PR)
5. **Drift check `last_pk <= 0`** — can't distinguish "fresh install" from "DB load failure"; silently skips validation on error
6. **Drift check race window** — consider re-loading last page before drift check as belt-and-suspenders
7. Lines 1562/1563 are f-strings with no interpolation (nit)

john-westcott-iv (approved):
All review feedback addressed. Crash recovery is correct (per-page advance with ascending ordering), drift detection now uses in-memory `last_advanced_pk` instead of a DB re-read, terminology is clear (watermark vs pagination token), and `order_by` redundancy is documented.

One minor CI note: `ruff format` is failing on a line-length preference in the error message — trivial fix.

LGTM.

PR #143, opened 2026-06-29
[AAP-80609] Optimize role assignment migration to use bulk queries

john-westcott-iv:
Requires: https://github.com/ansible/django-ansible-base/pull/1047

## Summary

Optimize `migrate_service_data` role assignment migration from O(N*3) queries to O(pages*~10), reducing migration time from hours to minutes for large installations (340k+ assignments).

### Bug Fixes
- **Fix merge flag warnings firing on every run**: `--merge-teams` and `--merge-organizations` defaulted to `True`, so argparse always populated them and the deprecation warning fired even when the user never passed the flag. Changed defaults to `None` and check with `is not None`.
- **Move `delete_legacy_authenticators` out of per-service loop**: This method cleans up gateway-local authenticator records and is idempotent. It was called once per service (3x as no-op) but only needs to run once per invocation.

### New Features
- **Add `--rerun` flag**: Bypasses the `has_migration_completed()` check so operators can re-run the command without manually resetting the migration flag via shell_plus. Useful for testing and benchmarking.

### Performance Optimizations
- **Bulk role assignment creation**: Replace per-assignment DB lookups and `give_permission` calls with batch processing — collect all unique role names, actor IDs, and object IDs from each page and resolve them in 3 bulk queries instead of 3+ per assignment. Bulk-create ObjectRoles and assignments via `bulk_create` with `ignore_conflicts`.
- **Deferred RBAC cache rebuild**: Rebuild `compute_team_member_roles()` and `compute_object_role_permissions()` once at the end (in a `finally` block) instead of per-assignment.
- **id__gt sliding window pagination**: Replace page-number pagination with `id__gt` sliding window. Page-number pagination is unstable when items are deleted mid-run (offsets shift, items get skipped). The sliding window always requests items after the last processed PK, which is stable against deletions and doubles as crash recovery.
- **Replace `get_new_resource_name` loop with single query**: The while loop called `.filter().exists()` per iteration to find a unique name suffix. Replaced with a single query using `order_by/first` to find the highest existing suffix.
- **DB-side service ordering**: Replace Python-side dict reordering with `Case/When` so the DB returns services in the correct order directly.

### Code Cleanup
- **Remove dead `merge_option` code**: The `merge` key was set in `resource_types_to_migrate` and copied into `resource_context` as `merge_option` but never read. Removed the dead keys and consolidated the declaration into the `OrderedDict` constructor.
- **Clean up `_report_migration_summary`**: Move `_ensure_superuser_consistency` and `mark_migration_completed` out of the reporting method into `handle` where they belong. Use `len(service_apis)` consistently instead of recomputing from success + failure counts.

## Dependencies
- Requires PR #151 (code refactor) — merged ✅
- Requires PR #152 (test refactor) — in review

## Test plan
- [ ] Existing test suite passes
- [ ] Benchmark with 100k role assignments shows significant speedup
- [ ] `--rerun` flag bypasses migration-completed check
- [ ] Merge flag warnings only fire when flags are explicitly passed
- [ ] Legacy authenticator cleanup runs once, not per-service

## Summary by CodeRabbit

* **New Features**
  * Added a safer, resumable service migration run with a `--rerun` option and clearer progress reporting.
  * Migration now processes services in a predictable order and finishes with a summary of successes and failures.
  * Role assignments are migrated in bulk for faster, more reliable runs.

* **Bug Fixes**
  * Improved handling of duplicate names, partially migrated users, and stale service IDs.
  * Legacy authenticator cleanup and superuser syncing now behave more consistently across services.

* **Tests**
  * Expanded automated coverage for logging, migration flows, role assignments, user merging, and superuser synchronization.

### DIFF
--- a/aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
@@ -46,10 +46,11 @@ class CursorStore:
         " DO UPDATE SET last_pk = EXCLUDED.last_pk"
     )
 
-    def __init__(self, service_slug, assignment_type):
+    def __init__(self, service_slug, assignment_type, log_fn=None):
         """Load the cursor from DB, setting self.last_pk once (immutably)."""
         self.service_slug = service_slug
         self.assignment_type = assignment_type
+        self._log_fn = log_fn
         self.last_pk = self._load()
 
     def _ensure_table(self):
@@ -66,12 +67,11 @@ class CursorStore:
                 row = cur.fetchone()
                 return row[0] if row else 0
         except Exception:
-            logger.warning(
-                "Failed to load cursor for %s/%s, will reprocess all assignments",
-                self.service_slug,
-                self.assignment_type,
-                exc_info=True,
-            )
+            msg = f"Failed to load cursor for {self.service_slug}/{self.assignment_type}, will reprocess all assignments"
+            if self._log_fn:
+                self._log_fn(msg, logging.WARNING)
+            else:
+                logger.warning(msg, exc_info=True)
             return 0
 
     def advance(self, pk):
@@ -81,16 +81,14 @@ class CursorStore:
         id__gt filter remains consistent across all pages of a run.
         If the upsert fails, a warning is logged but the run continues --
         the next invocation will reprocess from the old position, which
-        is safe because give_permission is idempotent.
+        is safe because bulk_create with ignore_conflicts is idempotent.
         """
         try:
             with connection.cursor() as cur:
                 cur.execute(self._SQL_UPSERT, [self.service_slug, self.assignment_type, pk])
         except Exception:
-            logger.warning(
-                "Failed to advance cursor for %s/%s to %d; next run will reprocess from the old position",
-                self.service_slug,
-                self.assignment_type,
-                pk,
-                exc_info=True,
-            )
+            msg = f"Failed to advance cursor for {self.service_slug}/{self.assignment_type} to {pk}; next run will reprocess from the old position"
+            if self._log_fn:
+                self._log_fn(msg, logging.WARNING)
+            else:
+                logger.warning(msg, exc_info=True)
--- a/aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
@@ -1,4 +1,5 @@
 import logging
+import traceback
 
 from django.db import connection
 
@@ -71,7 +72,7 @@ class CursorStore:
         except Exception:
             msg = f"Failed to load cursor for {self.service_slug}/{self.assignment_type}, will reprocess all assignments"
             if self._log_fn:
-                self._log_fn(msg, logging.WARNING)
+                self._log_fn(f"{msg}\n{traceback.format_exc()}", logging.WARNING)
             else:
                 logger.warning(msg, exc_info=True)
             return 0
@@ -91,6 +92,6 @@ class CursorStore:
         except Exception:
             msg = f"Failed to advance cursor for {self.service_slug}/{self.assignment_type} to {pk}; next run will reprocess from the old position"
             if self._log_fn:
-                self._log_fn(msg, logging.WARNING)
+                self._log_fn(f"{msg}\n{traceback.format_exc()}", logging.WARNING)
             else:
                 logger.warning(msg, exc_info=True)
--- a/aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
@@ -51,6 +51,8 @@ class CursorStore:
         self.service_slug = service_slug
         self.assignment_type = assignment_type
         self._log_fn = log_fn
+        # Set once at init; not updated by advance(). Do not read mid-run
+        # as a DB failure during advance() leaves this stale at 0.
         self.last_pk = self._load()
 
     def _ensure_table(self):
--- a/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
@@ -1,11 +1,11 @@
 import logging
-from typing import Any, Dict, List, Optional, Tuple, Type
+from typing import Any, Dict, List, Optional, Tuple
 
 from ansible_base.resource_registry.constants import SHARED_USER_RESOURCE_TYPE
 from ansible_base.resource_registry.models import Resource, service_id
 from ansible_base.resource_registry.rest_client import ResourceRequestBody
 from django.conf import settings
-from django.db import models, transaction
+from django.db import transaction
 
 
 class ResourceMigrationMixin:
@@ -21,53 +21,6 @@ class ResourceMigrationMixin:
             if data["count"] > 0:
                 return False
         return True
-
-    def get_new_resource_name(
-        self,
-        name: str,
-        unique_filter_kwargs: Dict[str, Any],
-        local_resource_model: Type[models.Model],
-        resource_type_name_field: str,
-        service_slug: str,
-    ) -> str:
-        """
-        Generate a unique name for a resource that doesn't conflict with existing resources.
-
-        When a resource name conflicts with an existing resource in the gateway, this method
-        generates a new name by prefixing with the service slug and adding a numeric suffix
-        if needed to ensure uniqueness.
-
-        Args:
-            name: Original resource name from upstream service
-            unique_filter_kwargs: Filter parameters used to check uniqueness
-            local_resource_model: Django model class for the resource type
-            resource_type_name_field: Field name used for the resource name
-
-        Returns:
-            A unique name that doesn't conflict with existing resources
-
-        Example:
-            If 'my-org' exists, will return 'service_my-org' or 'service_my-org1'
-        """
-        original_name = f'{service_slug}_{name}'
-
-        filter_kwargs = unique_filter_kwargs.copy()
-        filter_kwargs[f"{resource_type_name_field}__startswith"] = original_name
-        del filter_kwargs[resource_type_name_field]
-
-        highest = (
-            local_resource_model.objects.filter(**filter_kwargs)
-            .values_list(resource_type_name_field, flat=True)
-            .order_by(f"-{resource_type_name_field}")
-            .first()
-        )
-
-        if highest is None:
-            return original_name
-
-        suffix = highest[len(original_name) :]
-        counter = (int(suffix) + 1) if suffix.isdigit() else 1
-        return f"{original_name}{counter}"
 
     def update_resource_data(self, resource_type_name: str, original_resource_data: Any) -> Optional[Dict[str, Any]]:
         """
--- a/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
@@ -139,41 +139,10 @@ class ResourceMigrationMixin:
 
         return updated_resource_data
 
-    def _initialize_resource_sync_payloads(self, upstream_resource: Dict[str, Any], user_partial_migration: bool) -> Tuple[Dict[str, Any], Dict[str, Any]]:
-        """
-        Prepare payloads for creating Gateway resources and updating upstream resources.
-
-        This method sets up the data structures needed to:
-        1. Create a new resource in the gateway
-        2. Update the corresponding resource in the upstream service
-
-        For partially migrated users, the gateway resource retains the upstream
-        service_id to indicate it's not fully migrated.
-
-        Args:
-            upstream_resource: Complete resource data from upstream service
-            user_partial_migration: True if this is a user being partially migrated
-
-        Returns:
-            Tuple of (resource_creation_kwargs, updated_service_resource)
-            - resource_creation_kwargs: Data for creating Gateway resource
-            - updated_service_resource: Data for updating upstream resource
-        """
-        resource_creation_kwargs = {}
-        updated_service_resource = {}
-
-        resource_creation_kwargs["ansible_id"] = upstream_resource["ansible_id"]
-
-        if user_partial_migration:
-            # We do not update the service_id of a user on the service, only mark is_partially_migrated to True to exclude it from the
-            # while loop in migrate_resource()
-            updated_service_resource["is_partially_migrated"] = True
-            # The resource to be created in Gateway needs to show as unmigrated by having the original service_id
-            resource_creation_kwargs["service_id"] = self.upstream_service_id
-        else:
-            # if current resource is not shared.user or user is not partially migrated, we update the 'service_id' to Gateway's service_id
-            updated_service_resource["service_id"] = str(service_id())
-
+    def _initialize_resource_sync_payloads(self, upstream_resource: Dict[str, Any]) -> Tuple[Dict[str, Any], Dict[str, Any]]:
+        """Prepare payloads for creating Gateway resources and updating upstream resources."""
+        resource_creation_kwargs = {"ansible_id": upstream_resource["ansible_id"]}
+        updated_service_resource = {"service_id": str(service_id())}
         return resource_creation_kwargs, updated_service_resource
 
     def _reconcile_existing_resource(
@@ -317,11 +286,7 @@ class ResourceMigrationMixin:
             # Re-validate after potential superuser flag changes
             validated_resource_data = self._deserialize_and_validate_resource_data(upstream_resource, resource_context["type_serializer"])
 
-        # 'shared.user' type is treated differently
-        # If the user being migrated is not the current user (admin user), we need to check if we should partially migrate the user
-        user_partial_migration = False
-
-        resource_creation_kwargs, updated_service_resource = self._initialize_resource_sync_payloads(upstream_resource, user_partial_migration)
+        resource_creation_kwargs, updated_service_resource = self._initialize_resource_sync_payloads(upstream_resource)
 
         # handles case with existing resource and figure out if we should create a new resource in gateway or not
         create_gateway_resource = self._reconcile_existing_resource(upstream_resource, resource_context, validated_resource_data, updated_service_resource)
--- a/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
@@ -50,18 +50,24 @@ class ResourceMigrationMixin:
             If 'my-org' exists, will return 'service_my-org' or 'service_my-org1'
         """
         original_name = f'{service_slug}_{name}'
-        name = original_name
 
         filter_kwargs = unique_filter_kwargs.copy()
-        filter_kwargs[resource_type_name_field] = name
+        filter_kwargs[f"{resource_type_name_field}__startswith"] = original_name
+        del filter_kwargs[resource_type_name_field]
 
-        counter = 1
-        while local_resource_model.objects.filter(**filter_kwargs).exists():
-            name = original_name + str(counter)
-            filter_kwargs[resource_type_name_field] = name
-            counter += 1
+        highest = (
+            local_resource_model.objects.filter(**filter_kwargs)
+            .values_list(resource_type_name_field, flat=True)
+            .order_by(f"-{resource_type_name_field}")
+            .first()
+        )
 
-        return name
+        if highest is None:
+            return original_name
+
+        suffix = highest[len(original_name) :]
+        counter = (int(suffix) + 1) if suffix.isdigit() else 1
+        return f"{original_name}{counter}"
 
     def update_resource_data(self, resource_type_name: str, original_resource_data: Any) -> Optional[Dict[str, Any]]:
         """
@@ -374,7 +380,6 @@ class ResourceMigrationMixin:
             "type_serializer": resource_type.serializer_class,
             "type_name_field": resource_type.get_resource_config().name_field,
             "unique_fields": self.resource_types_to_migrate[resource_type_name]["unique_fields"],
-            "merge_option": self.resource_types_to_migrate[resource_type_name]["merge"],
             "LocalResourceModel": resource_type.content_type.model_class(),
         }
 
--- a/aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
@@ -274,12 +274,13 @@ class RoleAssignmentsMixin:
                 all_object_roles.update(object_roles)
                 self._log(f"  {assignment_type}: {created} assignments created", logging.INFO)
         finally:
-            if all_object_roles:
+            if total_created:
                 self._log(
-                    f"Rebuilding RBAC cache for {len(all_object_roles)} object roles",
+                    f"Rebuilding RBAC cache ({total_created} assignments created, {len(all_object_roles)} object roles)",
                     logging.INFO,
                 )
                 compute_team_member_roles()
-                compute_object_role_permissions(object_roles=all_object_roles)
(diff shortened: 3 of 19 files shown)


Dostonbek1 (approved):
## Code Review Summary

Excellent performance optimization that replaces per-assignment DB queries with batch operations, reducing total queries from ~600k to ~1.5k for large deployments (200k+ role assignments). The implementation is well-decomposed, follows DAB's existing patterns, and the deferred cache rebuild in a `finally` block is a correct safety measure.

**Files Reviewed**: 2 files
**Comments Posted**: 0 inline comments

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 0 code quality suggestions

### Overall Assessment
LGTM — approving.

### General Feedback
- The `_SKIP_SENTINEL` pattern for distinguishing "global assignment (None)" from "skip this item" is clean and well-documented.
- Using exact `Q()` tuple matching instead of three independent `__in` filters correctly avoids the cross-product blowup — good catch in the second commit.
- The `content_type__model` filter on actor Resource lookups adds defense-in-depth against theoretical PK collisions.
- The `finally` block for `compute_team_member_roles()` + `compute_object_role_permissions()` ensures partial assignments still get their cache rebuilt, which is critical for correctness on failure recovery.
- Test coverage is thorough: bulk ObjectRole sharing, idempotency, cache rebuild on partial failure, stale FK detection via bulk existence checks.
- The known limitation around global assignment idempotency (PostgreSQL NULL != NULL for unique constraints) is documented in the test comments, and the cursor mechanism limits the blast radius to at most one page on crash recovery — acceptable trade-off for the performance gain.
- All CI checks pass (unit tests, lint, SonarCloud, ATF integration tests, code coverage).

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1563,144 +1568,312 @@ def _paginate_and_create(self, list_fn, assignment_type: str, roles_to_exclude:
...
+                for r in Resource.objects.filter(ansible_id__in=actor_ansible_ids)
+            }
+
+        obj_all_ansible_ids = org_ansible_ids | team_ansible_ids
+        obj_resource_map: Dict[str, Resource] = {}
+        if obj_all_ansible_ids:
+            obj_resource_map = {
+                str(r.ansible_id): r
+                for r in Resource.objects.filter(ansible_id__in=obj_all_ansible_ids)
+            }
```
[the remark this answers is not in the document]
john-westcott-iv: Acknowledged. The ansible_id fields are UUIDs so PK collisions are effectively impossible, but this is a fair defense-in-depth suggestion. We'll consider adding content_type filtering in a follow-up if the security review requires it. For now the bulk lookup by ansible_id is correct and well-tested.

aap_gateway_api/management/commands/migrate_service_data.py
```diff
@@ -1719,33 +1892,37 @@ def migrate_role_assignments(self, service_slug: str, service_type_name: str) ->
...
+                if self._check_for_drift(list_fn, assignment_type, service_slug):
+                    drift_detected = True
+        finally:
+            if all_object_roles:
+                self._log(
+                    f"Rebuilding RBAC cache for {len(all_object_roles)} object roles",
+                    logging.INFO,
+                )
+                compute_team_member_roles()
+                compute_object_role_permissions(object_roles=all_object_roles)
```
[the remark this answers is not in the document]
john-westcott-iv: Fixed in fcc0763b. The finally block now gates on `total_created` (any assignments created) instead of `all_object_roles` (only object-scoped). This ensures `compute_team_member_roles()` runs even for global-only team assignments. `compute_object_role_permissions()` is still gated on `all_object_roles` since it needs the ObjectRole instances.

aap_gateway_api/tests/management/test_migrate_service_data_role_assignments.py
```diff
@@ -472,14 +476,72 @@ def test_no_drift_completes_normally(self):
...
+        with (
+            patch.object(cmd, "_paginate_and_create", return_value=(5, {mock_or})),
+            patch.object(cmd, "_check_for_drift", return_value=False),
+            patch("aap_gateway_api.management.commands.migrate_service_data.compute_team_member_roles") as mock_tmr,
+            patch("aap_gateway_api.management.commands.migrate_service_data.compute_object_role_permissions") as mock_corp,
+        ):
+            cmd.migrate_role_assignments("controller", "controller")
+
+        mock_tmr.assert_called_once_with()
+        mock_corp.assert_called_once_with(object_roles={mock_or})
```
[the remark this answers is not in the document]
john-westcott-iv: Fixed in fcc0763b. Test now uses distinct mocks (`mock_user_or`, `mock_team_or`) via `side_effect` and asserts the union `{mock_user_or, mock_team_or}` is passed to `compute_object_role_permissions`.

aap_gateway_api/tests/management/test_migrate_service_data_role_assignments.py
```diff
@@ -540,284 +602,309 @@ def test_check_for_drift_returns_false_on_api_error(self):
...
-        assert result is False
+        cmd = self._make_cmd()
+        results = [_make_remote_assignment("user", str(user.resource.ansible_id), "Test Bulk Idempotent Role")]
+
+        created1, _ = cmd._create_page_assignments(results, "user")
+        created2, _ = cmd._create_page_assignments(results, "user")
+        assert created1 == 1
+        # Second run: bulk_create with ignore_conflicts returns 0 (or 1
+        # depending on DB backend behavior — the key point is no error)
+        assert created2 >= 0
```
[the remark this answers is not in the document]
john-westcott-iv: Fixed in fcc0763b. The idempotency test now asserts `RoleUserAssignment.objects.filter(user__username='bulk-idempotent-user').count() == 1` after the second call, proving no duplicate rows are created.

jeffh-oss (approved):

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -50,18 +50,24 @@ def get_new_resource_name(
...
-        filter_kwargs[resource_type_name_field] = name
+        filter_kwargs[f"{resource_type_name_field}__startswith"] = original_name
+        del filter_kwargs[resource_type_name_field]
+
+        highest = (
+            local_resource_model.objects.filter(**filter_kwargs)
+            .values_list(resource_type_name_field, flat=True)
+            .order_by(f"-{resource_type_name_field}")
+            .first()
+        )
```
jay-steurer: **Critical: Lexicographic ordering produces wrong counter for multi-digit suffixes**

`order_by(f"-{resource_type_name_field}")` uses string sorting, not numeric. When suffixes exceed single digits, `"name9"` sorts higher than `"name10"` (because `'9' > '1'` character-wise).

Example: if `hub_myorg` through `hub_myorg10` all exist, this query returns `hub_myorg9` as "highest", computes `suffix="9"`, `counter=10`, and returns `hub_myorg10` — which already exists. This triggers an IntegrityError.

The old while-loop was immune because it checked existence at each step.

Suggested fix — fetch all matching names and compute the numeric max in Python:
```python
existing = list(
    local_resource_model.objects.filter(**filter_kwargs)
    .values_list(resource_type_name_field, flat=True)
)
if not existing:
    return original_name
max_counter = 0
for name in existing:
    s = name[len(original_name):]
    if s == '':
        max_counter = max(max_counter, 0)
    elif s.isdigit():
        max_counter = max(max_counter, int(s))
return f"{original_name}{max_counter + 1}" if max_counter > 0 or original_name in existing else original_name
```

Additionally, `__startswith` can match unrelated names (e.g., `service_my-organic` matches `service_my-org`). Consider a more precise filter.

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
```diff
@@ -45,277 +38,258 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
...
+            if total:
+                self._log_progress(progress_label, created, total)
 
-            if not data.get('next'):
+            last_pk_on_page = results[-1].get('id')
+            if last_pk_on_page is not None:
+                cursor.advance(last_pk_on_page)
+                base_filters['id__gt'] = str(last_pk_on_page)
+            else:
                 break
```
jay-steurer: **Critical: Missing `id` field silently truncates pagination**

The old code raised `RuntimeError("API returned assignment without 'id' field -- cannot advance cursor")` — an explicit signal of an incompatible upstream DAB version. The new code silently `break`s after processing only the current page.

If the upstream service is running an incompatible version that omits `id` from responses, this silently drops all remaining pages of assignments. The migration reports success with a partial count, and the operator has no indication that data was lost.

Suggested fix — restore the error, or at minimum log a prominent warning:
```python
else:
    self._log(
        f"Warning: API returned {assignment_type} assignment without 'id' field. "
        f"Stopping pagination early -- remaining assignments will not be processed. "
        f"Ensure the upstream service is running a compatible DAB version.",
        logging.WARNING,
    )
    break
```

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
```diff
@@ -45,277 +38,258 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
...
+                **{actor_field: actor_pk},
+                object_role=None,
+                role_definition_id=rd.id,
+                content_type_id=None,
+                object_id=None,
+            )
+            for actor_pk, rd, _, _, _ in assignments
+        ]
+
+        result = AssignmentModel.objects.bulk_create(assignment_objs, ignore_conflicts=True)
```
jay-steurer, line 240: **Critical: Global assignments can duplicate on crash recovery**

`bulk_create(ignore_conflicts=True)` only deduplicates when there's a unique constraint to trigger a conflict. Global assignments (where `content_type_id`, `object_id`, and `object_role` are all `None`) may lack such a constraint.

If the process crashes between `bulk_create` returning and `cursor.advance` persisting, the next run reprocesses the same page. Object-scoped assignments are safe (unique constraint prevents duplicates), but global assignments would be **duplicated**. The old `give_global_permission` used `get_or_create` internally which was inherently idempotent.

Suggested fix — filter out existing global assignments before `bulk_create`:
```python
existing = set(
    AssignmentModel.objects.filter(
        object_role=None, content_type_id=None, object_id=None
    ).values_list(actor_field, 'role_definition_id')
)
new_objs = [
    obj for obj, (actor_pk, rd, _, _, _) in zip(assignment_objs, assignments)
    if (actor_pk, rd.id) not in existing
]
if new_objs:
    AssignmentModel.objects.bulk_create(new_objs, ignore_conflicts=True)
return len(new_objs)
```

---

**Important: `len(result)` returns count submitted, not count inserted**

With `ignore_conflicts=True`, Django's `bulk_create` returns all objects in the input list regardless of how many were actually inserted. This inflates the "created" count in operator logs. On `--rerun`, the operator sees e.g. "5000 assignments created" when 0 new rows were inserted. Also causes unnecessary RBAC cache rebuilds since `total_created` is always truthy.

Consider querying count before/after for the actual delta, or at minimum rename to "processed".

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
```diff
@@ -45,277 +38,258 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
...
+                    object_id=object_role.object_id,
+                )
+            )
 
-        After all pages are processed, a post-run drift check detects
-        assignments created during the run.  If drift is found, RuntimeError
-        is raised so the installer retries and the cursor picks up the
-        new items.
+        result = AssignmentModel.objects.bulk_create(assignment_objs, ignore_conflicts=True)
+        return len(result), set(or_map.values())
```
jay-steurer, line 287: Same `len(result)` inflation issue as `_bulk_create_global_assignments` — with `ignore_conflicts=True`, this returns the count submitted, not the count actually inserted.

aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
```diff
@@ -66,12 +69,11 @@ def _load(self) -> int:
...
         except Exception:
-            logger.warning(
-                "Failed to load cursor for %s/%s, will reprocess all assignments",
-                self.service_slug,
-                self.assignment_type,
-                exc_info=True,
-            )
+            msg = f"Failed to load cursor for {self.service_slug}/{self.assignment_type}, will reprocess all assignments"
+            if self._log_fn:
+                self._log_fn(msg, logging.WARNING)
```
jay-steurer: **Important: Exception tracebacks are lost when `log_fn` is provided**

The old code always used `logger.warning(msg, exc_info=True)` which included the full traceback. The `log_fn` path only passes the message text — the exception type, message, and traceback are silently discarded.

Since `migrate_role_assignments` now always passes `log_fn=self._log`, the `exc_info=True` fallback path is effectively dead code in production. DB connection failures, permission errors, and other cursor-related exceptions lose all diagnostic information.

Suggested fix:
```python
import traceback
if self._log_fn:
    self._log_fn(f"{msg}\n{traceback.format_exc()}", logging.WARNING)
else:
    logger.warning(msg, exc_info=True)
```

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
```diff
@@ -45,277 +38,258 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
...
-            )
-        return False
+            for assignment_type in ('user', 'team'):
+                list_fn = self.client.list_user_assignments if assignment_type == 'user' else self.client.list_team_assignments
+                cursor = CursorStore(service_slug, assignment_type, log_fn=self._log)
+                created, object_roles = self._paginate_and_create(list_fn, assignment_type, roles_to_exclude, cursor)
+                total_created += created
+                all_object_roles.update(object_roles)
+                self._log(f"  {assignment_type}: {created} assignments created", logging.INFO)
+        finally:
```
jay-steurer, line 315: **Important: `finally` block can mask the original exception**

If `_paginate_and_create` raises (e.g., HTTP 500 from the team assignment API) and then `compute_team_member_roles()` in the `finally` block also raises (e.g., DB connection lost), the original exception is **replaced** by the finally-block exception. The operator sees a cache rebuild error instead of the actual root cause.

Also, if user assignments succeed but team assignments fail, the cache is rebuilt with only user-assignment object roles — an incomplete/inconsistent state.

Suggested fix — wrap the cache rebuild in its own try/except:
```python
finally:
    if total_created:
        try:
            self._log(f"Rebuilding RBAC cache ...", logging.INFO)
            compute_team_member_roles()
            if all_object_roles:
                compute_object_role_permissions(object_roles=all_object_roles)
        except Exception:
            self._log(
                "RBAC cache rebuild failed; cache may be stale until next successful run",
                logging.ERROR,
            )
```

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
```diff
@@ -45,277 +38,258 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
...
-            base_filters['id__gt'] = str(cursor.last_pk)
+
+        progress_label = f"{assignment_type} role assignments"
 
         while True:
-            response = list_fn(filters={**base_filters, 'page': page})
+            response = list_fn(filters=base_filters)
             if response.status_code != 200:
-                self._raise_fetch_error(response, assignment_type, page)
+                self._raise_fetch_error(response, assignment_type, 0)
```
jay-steurer, line 66: **Minor: Page number hardcoded to `0` reduces diagnostic value**

Since page-number pagination was replaced with `id__gt` sliding window, the `page` parameter to `_raise_fetch_error` is now always `0`. The error message "Failed to fetch user assignments page 0: HTTP 500" gives operators no information about how far the migration progressed.

Consider passing `base_filters.get('id__gt', '0')` instead and updating the error message template to say "after id" rather than "page".

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
```diff
@@ -45,277 +38,258 @@ def _raise_fetch_error(response, assignment_type: str, page: int) -> None:
...
-        After all pages are processed, a post-run drift check detects
-        assignments created during the run.  If drift is found, RuntimeError
-        is raised so the installer retries and the cursor picks up the
-        new items.
+        result = AssignmentModel.objects.bulk_create(assignment_objs, ignore_conflicts=True)
+        return len(result), set(or_map.values())
 
-        give_permission is idempotent (uses get_or_create), so replaying
-        a partial page after a crash is safe.
+    def migrate_role_assignments(self, service_slug: str, service_type_name: str) -> None:
```
jay-steurer, line 289: **Note: Drift detection was removed entirely**

The old `_check_for_drift` method detected assignments created on the upstream service *during* the migration run (a single `page_size=1` API call per type — negligible overhead). Its removal means assignments created in the window between "last page fetched" and "migration flag set" are silently lost with no warning.

The sliding window handles *deletions* correctly, but drift detection was about *insertions*. Consider restoring a lightweight warning-only check:
```python
check_resp = list_fn(filters={'order_by': 'id', 'id__gt': str(cursor.last_pk), 'page_size': '1'})
if check_resp.status_code == 200 and check_resp.json().get('count', 0) > 0:
    self._log(f"Warning: new {assignment_type} assignments appeared during migration", logging.WARNING)
```

jay-steurer (commented):
## PR Review Summary

Strong performance optimization overall — the bulk query approach is well-architected and the test coverage is thorough. Found 3 critical issues (correctness bugs), 5 important issues (error handling / diagnostics), and 2 suggestions. See inline comments for details.

### Critical Issues
1. `get_new_resource_name` lexicographic ordering produces incorrect counter for multi-digit suffixes
2. Missing `id` field silently truncates pagination instead of raising
3. Global role assignments can duplicate on crash recovery (`ignore_conflicts` requires a unique constraint)

### Important Issues
4. `bulk_create` + `ignore_conflicts` inflates the reported "created" count
5. CursorStore `log_fn` path loses exception tracebacks
6. RBAC cache `finally` block can mask the original exception
7. Removed drift detection loses visibility into concurrent modifications
8. Page number hardcoded to `0` in error messages

### Suggestions
9. `--rerun` flag has no behavioral test (only parser registration is tested)
10. `_bulk_create_object_assignments` ObjectRole-not-found path is untested

### Strengths
- Sliding window pagination thoroughly tested including mid-run deletion stability
- RBAC cache `finally` block has excellent coverage (4 test scenarios including exception path)
- Performance tests validate O(1) query complexity per page
- `_resolve_single_assignment` has complete branch coverage
- Dead code cleanup (`merge_option`, `user_partial_migration`) is clean

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
zkayyali812: I don't see this function being used anywhere. Is this dead code now?

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
john-westcott-iv: Good catch, but this method is actually dead code — its only caller was removed in commit 18801eb4 (AAP-54038, Oct 2025). Removed the method and its tests in 24730690.

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
john-westcott-iv: Fixed in 24730690. Restored the RuntimeError — if the API doesn't include an `id` field, something is fundamentally broken and we should fail loudly.

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
john-westcott-iv: Fixed in 24730690. Global assignments now query for existing `(actor, role_definition)` combos before `bulk_create` to prevent duplicates, since PostgreSQL treats `NULL != NULL` in unique constraints.

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
john-westcott-iv: Acknowledged. `len(result)` from `bulk_create(ignore_conflicts=True)` returns the input list length, not actual inserts. This is cosmetic — the count is used only for logging, and the actual DB state is correct. Not changing this as the Django API doesn't provide actual insert counts.

aap_gateway_api/management/commands/_migrate_service_data/cursor_store.py
john-westcott-iv: Fixed in 24730690. The `log_fn` path now includes `traceback.format_exc()` in the message so the full exception context is preserved.

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
john-westcott-iv: Fixed in 24730690. Cache rebuild in the `finally` block is now wrapped in `try/except` — if it fails, we log a warning instead of masking the original exception.

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
john-westcott-iv: Intentional. With the id__gt sliding window, there's no page number — just a cursor position. The `0` is the page parameter to `_raise_fetch_error` which is a holdover from the old page-based code. The log message includes the HTTP status and response body which are the diagnostically useful parts. The cursor's current `id__gt` value is visible in the preceding API call logs.

aap_gateway_api/management/commands/_migrate_service_data/role_assignments.py
john-westcott-iv: Agreed — drift detection was intentionally removed. Added a docstring note in 24730690 explaining the rationale: the id__gt sliding window handles deletions, and new assignments get higher PKs so they appear on subsequent pages. Additionally, by the time `migrate_service_data` runs, all services are already on 2.6 with reverse sync active, so any net-new items sync to gateway automatically.

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
john-westcott-iv: Confirmed dead code — the only production caller was removed in commit 18801eb4 (AAP-54038, Oct 2025). Removed the method and its tests in 24730690.

bhavenst (approved):
Approving — no critical or important issues found in my review. Approval is contingent on any outstanding comments from other reviewers being addressed.

**Minor observations (not blocking):**

- `_bulk_create_object_assignments`: `len(result)` from `bulk_create(ignore_conflicts=True)` counts objects *passed*, not *inserted* — logged "created" count could be slightly inflated on re-runs. Data correctness unaffected.
- RBAC cache rebuild `except Exception` handler doesn't include traceback (`exc_info`), which could make debugging harder if the rebuild fails. Low impact since the cache self-heals on next request.

Overall the bulk resolution approach is clean, the `finally`-block RBAC rebuild is well-tested, and the new perf tests will catch scaling regressions.

jay-steurer (approved):
Reviewed the fix commit (2473069) addressing the earlier feedback. All six issues are properly resolved:

- **RuntimeError restored** for missing `id` field — propagates correctly, no surrounding catch that would swallow it.
- **Global assignment dedup** — pre-filter with exact tuple matching is correct. Theoretical TOCTOU race exists but is not a realistic concern for a migration command.
- **Traceback preservation** — `traceback.format_exc()` in the except handler is correct; `_log` splits on newlines for structured logging.
- **Finally block masking** — inner try/except correctly isolates cache rebuild failures from the original exception.
- **Drift detection rationale** — id__gt sliding window handles this naturally; docstring explains the sub-second gap on the final page.
- **Dead code removal** — clean, no remaining references.

The `len(result)` cosmetic count from `bulk_create(ignore_conflicts=True)` is acknowledged and logging-only — fine as-is.

LGTM.

PR #148, opened 2026-07-02
[NO_JIRA] Fix SonarCloud file encoding warning by setting UTF-8 locale

tznamena:
## Description

- **What is being changed?** Sets `LC_ALL` and `LANG` environment variables to `en_US.UTF-8` on both SonarCloud Scan steps (PR analysis and branch analysis) in the CI workflow.
- **Why is this change needed?** SonarCloud is reporting a file encoding warning: _"There are problems with file encoding in the source code."_ The scanner's JVM relies on the system locale to determine its default charset. CI container runners often default to `LANG=C` (ASCII), which causes the scanner to fail decoding non-ASCII characters — even when `sonar.sourceEncoding=UTF-8` is set in `sonar-project.properties`.
- **How does this change address the issue?** By explicitly setting `LC_ALL` and `LANG` to `en_US.UTF-8` in the scanner step environment, the JVM's default charset matches the declared source encoding, and non-ASCII characters (em-dashes, checkmarks, arrows used in comments and log messages) are decoded correctly.

## Type of Change

- [x] Configuration change

## Self-Review Checklist

## Testing Instructions

This will be tested with the next PR that run SonarCloud Code Analysis.

### Investigation Summary

The codebase uses Unicode characters (em-dashes `—`, checkmarks `✓`, arrows `→` `↔`) in comments and log messages across 11+ files. This is an established pattern, not introduced by any single PR. The `sonar-project.properties` file correctly declares `sonar.sourceEncoding=UTF-8`, but the scanner also depends on the JVM's default charset, which is governed by the OS locale. The fix aligns the CI environment's locale with the declared encoding.

## Summary by CodeRabbit

* **Chores**
  * Updated the analysis workflow environment to use a consistent UTF-8 locale setting.
  * Applied the same environment configuration across both PR and branch analysis steps.


jay-steurer (approved):
LGTM, test failure is unrelated

PR #155, opened 2026-07-06
[AAP-80758] Fix org admin visibility of user role assignments on remote objects

aap-platform-services-cicd-bot-sa:
## Summary

Organization admins with `ORG_ADMINS_CAN_SEE_ALL_USERS` enabled cannot see or manage user role assignments on remote objects (e.g. AWX job templates, projects) for users outside their organization. This causes the "Manage roles" dialog to show "None" and the user's Roles tab to display a permission error. The fix adds a `filter_queryset` override to `GatewayRoleUserAssignmentViewSet` that bypasses the `visible_items` filter for org admins, mirroring the existing fix for team assignments (AAP-70503).

## Root Cause

`GatewayRoleUserAssignmentViewSet` was missing a `filter_queryset` override that was already implemented for `GatewayRoleTeamAssignmentViewSet` as part of AAP-70503. The gateway sets `ANSIBLE_BASE_ENFORCE_REMOTE_OBJECT_PERMISSIONS=False`, which makes the `RoleEvaluation` cache incomplete for remote objects. The `BaseAssignmentViewSet.filter_queryset` calls `visible_items()` which relies on this incomplete cache, causing user assignments on remote objects to be hidden from org admins.

## Changes

- `aap_gateway_api/views/api/v1/role.py`: Added `filter_queryset` method to `GatewayRoleUserAssignmentViewSet` that bypasses `visible_items` for users who `can_view_all_users` (superusers and org admins when `ORG_ADMINS_CAN_SEE_ALL_USERS` is enabled), identical to the pattern in `GatewayRoleTeamAssignmentViewSet`
- `aap_gateway_api/tests/views/api/v1/role/test_user_assignment_visibility.py`: Added regression tests covering:
  - Org admin sees cross-org user assignment on remote object when setting enabled
  - Org admin cannot see cross-org user assignment when setting disabled
  - Superuser always sees the assignment
  - Regular user cannot see the assignment
  - Org admin sees assignment in unfiltered list
  - Org admin sees local object assignment regardless of setting

## Risk Assessment

- **Confidence:** High
- **Risk:** Low -- the fix mirrors an existing, tested pattern (AAP-70503 team fix) and only affects visibility of assignments for org admins, not the assignment creation or deletion logic
- **Scope:** 2 file(s) changed

## Testing

- [ ] Unit tests pass
- [ ] Regression test added
- [ ] Lint checks pass

## JIRA

- Ticket: [AAP-80758](https://issues.redhat.com/browse/AAP-80758)

---
*Assisted-by: Debuggernaut (claude-opus-4-6) <noreply@redhat.com> | Review carefully before merging.*

[AAP-80758]: https://redhat.atlassian.net/browse/AAP-80758?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved visibility for role **user** assignments and **team** assignments on remote objects for organization administrators based on the **“allow organization administrators to see all users”** setting.
  * Superusers can always retrieve and list assignments; regular users still can’t see cross-org remote assignments.
  * Local-object assignments remain visible regardless of the setting.
  * List and retrieve endpoints now apply the same visibility rules and avoid duplicate results.
* **Tests**
  * Added coverage for privileged-bypass behavior, unfiltered list scenarios, and detail-view retrieval.


jay-steurer (commented):
## PR Review Summary

**Verdict: Approve** — Clean, well-structured fix that correctly mirrors the existing `GatewayRoleTeamAssignmentViewSet` pattern (AAP-70503). MRO bypass is sound, security boundary is correct, test coverage is thorough.

### Important (1)

- **No audit logging on the security filter bypass** (`role.py:170`): When `can_view_all_users` returns `True`, the `visible_items` filter is completely bypassed — the org admin sees *all* user role assignments across all orgs. There's no log entry recording who triggered the bypass or when. A `logger.debug(...)` when the bypass path is taken would help incident response. Same gap exists in the team assignment counterpart, so this is a pattern-level concern.

### Suggestions (4)

- **`user_orgb` fixture declares `org_b` dependency but never uses it** (`test_user_assignment_visibility.py:27-28`): Unlike the team test where `team_orgb` sets `organization=org_b`, this user has no org association. Either assign the user to `org_b` via an Organization Member role or drop the unused dependency.
- **Hardcoded `object_id=106` in test assertions**: Use `cross_org_user_assignment.object_id` instead of the literal `106` to make the connection explicit and prevent silent drift.
- **Missing test for the detail (retrieve) endpoint**: All 6 tests hit the list endpoint. DRF's `get_object()` also calls `filter_queryset`, so a test hitting `/role_user_assignments/<id>/` would guard against future divergence.
- **Fixture duplication with team test file**: 7 fixtures are nearly identical across user and team test files. Extracting into `role/conftest.py` would reduce duplication. Not blocking — follows the established pattern.

### Strengths

- Correct MRO bypass verified against upstream source
- Security boundary properly enforced via `can_view_all_users`
- Comprehensive test coverage (6 tests, 4 actor types, filtered + unfiltered, local + remote)
- Accurate, well-structured comments consistent with team viewset

john-westcott-iv:
## Code Review Notes

Code change looks good — mirrors the existing `GatewayRoleTeamAssignmentViewSet.filter_queryset` pattern (AAP-70503) exactly. CI passing, logic is correct.

### Edge cases to consider for test coverage

The current tests cover the core scenarios well. A few additional edge cases worth considering:

1. **Org admin sees their OWN assignment on a remote object** — the fixture creates an assignment for `user_orgb`, but doesn't test the case where the org admin themselves has a remote assignment. Verifies self-referencing visibility works.

2. **Assigned user in the SAME org as the org admin** — `user_orgb` isn't actually added to any org (fixture creates the user but no org membership). Worth testing what happens when the assigned user IS in the org admin's org — does `visible_items` already show it without the fix? If so, verify no duplicates.

3. **Detail endpoint (retrieve)** — all tests hit the list endpoint. Does the single-object retrieve (`roleuserassignment-detail`) also work correctly for the org admin? `filter_queryset` affects list, but `get_object` may have its own visibility check.

These are non-blocking — the existing coverage is solid for the bug being fixed.

john-westcott-iv (approved):
LGTM — mirrors the proven team assignment fix (AAP-70503). Edge case suggestions posted as a separate comment.

jay-steurer (approved):

PR #165, opened 2026-07-17
AAP-83321 - Add sync_service_ids command and lazy service_id fallback in auth

cshiels-ie:
## Problem

`migrate_service_data` uses a global completion flag. On the first upgrade (2.5→2.6) it runs, writes `service_id` onto each `ServiceCluster`, and marks itself done. On all subsequent upgrades it exits immediately. New services introduced in later releases (e.g. metrics-service in 2.7) are left with `service_id=null`, causing every JWT auth attempt from those services to fail with a 403.

## Solution

Two complementary layers sharing a single utility module (`service_id_sync.py`):

**1. `sync_service_ids` management command** — idempotent, safe on every upgrade

```bash
# Fill in any missing service_ids (normal upgrade path)
aap-gateway-manage sync_service_ids

# Re-fetch all IDs (e.g. after a service re-deployment)
aap-gateway-manage sync_service_ids --force
```

Individual service failures (503, unreachable) are reported as stderr warnings and the command exits 0 — expected when not all services are deployed. An unhandled exception (e.g. DB down) propagates as a non-zero exit.

**2. Lazy fallback in `validate_service_token()`** — safety net if the installer doesn't call the command. When `ServiceCluster.DoesNotExist` is raised, the auth path attempts to populate the missing `service_id` from the service's metadata endpoint before returning a 403.

- Restricted to known `DefaultServiceType` values (excluding GATEWAY) — unknown or custom service types are never probed during token validation.
- A 60-second per-issuer cooldown prevents repeated DB+HTTP calls from unknown or attacker-supplied tokens. After failed pruning the dict is capped to prevent unbounded growth under a token flood.
- At most 5 clusters are probed per request to bound outbound HTTP calls.
- All DB and HTTP operations inside the lazy path are wrapped in a broad `except Exception` so a transient DB error (e.g. `OperationalError`) returns `None` rather than propagating a 500 through the gRPC auth handler.
- One-time self-healing: once `service_id` is written, every subsequent request finds the cluster normally and the lazy path never fires again for that cluster.

## What changed

| File | Change |
|------|--------|
| `aap_gateway_api/utils/service_id_sync.py` | New shared utility — fetch, UUID-normalize, write `service_id`; N+1-free queries; per-issuer cooldown; enum-restricted and exception-guarded lazy path |
| `aap_gateway_api/management/commands/sync_service_ids.py` | New command — `--force` only; individual failures are warnings, not errors |
| `aap_gateway_api/utils/service_token.py` | Extract `_verify_token_signature` + `_get_user_from_payload` helpers; lazy populate returns cluster directly, eliminating an extra `get()` |
| `aap_gateway_api/tests/utils/test_service_id_sync.py` | Utility tests |
| `aap_gateway_api/tests/management/test_sync_service_ids.py` | Command tests |
| `aap_gateway_api/tests/authentication/test_service_token_auth.py` | Lazy path tests |

No changes to `migrate_service_data`, service orchestration, or model migrations.

## Design notes

- `populate_missing_service_ids` fetches all `ServiceAPIRoute`s for target clusters in one query (no N+1).
- `_fetch_service_id_for_route` normalises the returned value via `str(uuid.UUID(raw_id))` to canonical lowercase before saving, preventing case or format differences from blocking the `iss` comparison.
- Conditional `UPDATE` (`WHERE service_id IS NULL`) is concurrency-safe without a lock.
- `try_populate_service_id` uses `filter(__in=_LAZY_SYNCABLE_TYPES)` (not `exclude(GATEWAY)`) — only known service types are ever probed on the auth hot path.
- No `--username` flag — `GWResourceAPIClient` already calls `get_default_user()` (first superuser) when no user is supplied.

## AI Assistance
This change was developed with assistance from Claude Code (Claude Sonnet 4.6).
All generated code was reviewed, tested, and validated before merging.

## Summary by CodeRabbit

* **New Features**
  * Added a management command to synchronize missing service IDs, with optional forced refreshes.
  * Service IDs can now be populated automatically during token authentication when missing.
  * Added safeguards for invalid metadata, repeated lookup attempts, and unsupported service types.
  * Synchronization reports successful and failed cluster updates.

* **Bug Fixes**
  * Improved token validation for issuer mismatches and required token types.
  * Prevented failed refreshes from overwriting existing service IDs.

* **Tests**
  * Expanded coverage for synchronization, authentication fallback, cooldown behavior, and error handling.


zkayyali812 (commented):
Hi @cshiels-ie 
Im a bit curious on a few things -
1. If the original approach of migrate_service_data doesn't work, should we roll this back?
2. With the fallback, I expect we don't need any installer changes?
3. I'd like this change to work in a way that doesn't require future services to make jewel changes. This PR seems close, but can we make this more service agnostic?

cshiels-ie:
> Hi @cshiels-ie Im a bit curious on a few things -
> 
> 1. If the original approach of migrate_service_data doesn't work, should we roll this back?
> 2. With the fallback, I expect we don't need any installer changes?
> 3. I'd like this change to work in a way that doesn't require future services to make jewel changes. This PR seems close, but can we make this more service agnostic?

1. Currently I still need the change to migrate_service_data as it add metrics into the enum in `service_type.py` that currently controls access via gateway to the full RBAC system. 
2. Yes there is no installer or operator changes needed with the fallback option, and I made the command more like a ondemand command used for manual fixes rather than a required command this needs to be run at all new installs or upgrades. 
3. I think the missing piece is just registering the enum from `service_type.py` and the additional info, original I had it in the command but I felt there might be a need for further decision/discussion.

cshiels-ie:
Changing this to draft instead looking at just this change https://github.com/ansible/jewel/pull/166

PR #167, opened 2026-07-20
[AAP-82119] Batch resource migration using bulk-update endpoint

Dostonbek1:
Requires: https://github.com/ansible/django-ansible-base/pull/1054

## Description

- **What:** Refactor `migrate_resource()` to process entire API pages as batches instead of one item at a time, using the new DAB bulk-update endpoint.
- **Why:** Each resource triggers an individual HTTP PATCH to the upstream controller. For 3,962 users this means 3,962 sequential round-trips taking 203s (3.4 min). Role assignments (119,721 items) already use bulk optimization and take only 51s.
- **How:** New `_process_resource_page_batch()` method validates all items on the page, reconciles conflicts, bulk-creates gateway resources in a single transaction, then sends one `bulk_update_resources()` HTTP call per page.

## Type of Change

## Self-Review Checklist

### Prerequisites

- Full AAP environment with controller service running
- DAB must include the bulk-update endpoint (PR: https://github.com/ansible/django-ansible-base/pull/1054)

### Steps to Test

1. Deploy with both PRs merged
2. Run `manage.py migrate_service_data --username admin`
3. Verify resource migration (users, orgs, teams) completes successfully
4. At Kyndryl scale (~4k users, ~2k teams, ~130 orgs): total resource migration should complete under 2 minutes

### Expected Results

- Resources migrate correctly (same behavior as before)
- HTTP round-trips reduced from N (one per resource) to N/page_size (one per page)
- Transaction rollback on bulk-update failure (no partial state)
- Progress logging still reports per-page progress

### Required Actions

- [x] Blocked by PR/MR: https://github.com/ansible/django-ansible-base/pull/1054
  - DAB bulk-update endpoint must be merged first

### Performance Impact

| Metric | Before | After |
|--------|--------|-------|
| HTTP calls (3,962 users) | 3,962 | ~80 (page_size=50) |
| HTTP calls (2,288 teams) | 2,288 | ~46 |
| Projected user migration time | 203s | <10s |
| Projected total resource migration | ~7 min | <2 min |

---

## Summary by CodeRabbit

* **New Features**
  * Added bulk upstream resource updates during migration to reduce per-item API calls.
  * Improved migration flow with page/batch processing, configurable chunking, transient-error retries, and reconciliation-aware payloads.
  * Added stall detection to abort runs after repeated zero progress.
  * Added warnings when some upstream items fail to update while still progressing.

* **Bug Fixes**
  * Updated upstream migration payloads to use the `new_service_id` field name.

* **Tests**
  * Expanded unit coverage for batch processing, bulk update retries/error handling, payload generation, and stalled-migration circuit breakers.

### DIFF
--- a/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
@@ -1,9 +1,10 @@
 import logging
+import time
 from typing import Any, Dict, List, Optional, Tuple
 
+import requests
 from ansible_base.resource_registry.constants import SHARED_USER_RESOURCE_TYPE
 from ansible_base.resource_registry.models import Resource, service_id
-from ansible_base.resource_registry.rest_client import ResourceRequestBody
 from django.conf import settings
 from django.db import transaction
 
@@ -95,7 +96,7 @@ class ResourceMigrationMixin:
     def _initialize_resource_sync_payloads(self, upstream_resource: Dict[str, Any]) -> Tuple[Dict[str, Any], Dict[str, Any]]:
         """Prepare payloads for creating Gateway resources and updating upstream resources."""
         resource_creation_kwargs = {"ansible_id": upstream_resource["ansible_id"]}
-        updated_service_resource = {"service_id": str(service_id())}
+        updated_service_resource = {"new_service_id": str(service_id())}
         return resource_creation_kwargs, updated_service_resource
 
     def _reconcile_existing_resource(
@@ -149,7 +150,7 @@ class ResourceMigrationMixin:
         # so this will correct the service_id and possibly update the stale resource_data
         if str(existing_resource.ansible_id) == resource_ansible_id:
             create_gateway_resource = False
-            updated_service_resource["service_id"] = existing_resource.service_id
+            updated_service_resource["new_service_id"] = str(existing_resource.service_id)
 
             if incoming_data == local_data:
                 self._log(f"Correcting service_id of {resource_type.name} with name {upstream_resource['name']}.", logging.INFO)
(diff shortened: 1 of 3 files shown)


Dostonbek1:
## Testing Results: Production-Scale End-to-End Migration Benchmark

Tested against a local AAP dev instance (AAP 2.7, controller 4.8.3) with production-scale data:
- **3,962 users**
- **2,288 teams**
- **130 organizations**
- **57 role definitions**
- **6,428 total resources**

### Performance Results (Controller Service-Index Endpoint)

Benchmark simulates the exact path `migrate_service_data` takes: Gateway calls the controller's `POST /service-index/resources/bulk-update/` with batches of 50 items per page.

| Approach | Time | Rate | Speedup |
|----------|------|------|---------|
| **Batch processing (this PR)** | **9.2s** | 700 items/sec | **52x** |
| Per-item PATCH (old approach) | 481.2s (8.0 min) | 13.4 items/sec | 1x (baseline) |

### Comparison with Role Assignment Optimization (AAP-80609)

| Optimization | Items | Time | Speedup |
|---|---|---|---|
| Role assignments (AAP-80609) | 119,721 | 51s | 33x |
| **Resource migration (this PR)** | **6,428** | **9.2s** | **52x** |

### Platform Services Test Suite

Ran `make test-atf-run-ps` against the updated instance:
```
166 passed, 2 failed, 1 skipped, 22 deselected, 7 warnings
```
The 2 failures are pre-existing and unrelated (service_cluster PATCH proxy permissions).

### Acceptance Criteria

> Resource migration at production-scale should complete in under 2 minutes.

**Result: 9.2 seconds — well under target. ✅**

The full `migrate_service_data` command now completes resource migration 52x faster by batching page-level updates through the DAB bulk-update endpoint (PR ansible/django-ansible-base#1054).

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -253,6 +253,100 @@ def _process_and_migrate_resource_item(self, upstream_resource_item: Dict[str, A
...
+            bulk_item["resource_data"] = updated_service_resource["resource_data"]
+        return bulk_item
+
+    def _send_bulk_update(self, bulk_update_items: List[Dict[str, Any]]) -> int:
+        """Send bulk update to upstream and return the number of successfully updated items.
+
+        Per-item errors are logged as warnings. Items whose upstream update failed
+        will reappear on the next query (the filter naturally retries them).
+        """
+        resp = self.client.bulk_update_resources(bulk_update_items)
```
jay-steurer: **Critical: Unhandled network exceptions crash migration after local commits**

`self.client.bulk_update_resources()` can raise `ConnectionError`, `Timeout`, `SSLError`, or any `requests.exceptions.RequestException`. These are not caught, so they propagate up and crash `migrate_resource()`. Since local `Resource.create_resource()` calls have already committed (line 340-342), this leaves local resources orphaned until a manual re-run.

The old per-item code had the same exposure but wrapped the HTTP call inside `transaction.atomic()`, which would at least roll back the local create.

**Suggested fix:**
```python
def _send_bulk_update(self, bulk_update_items):
    try:
        resp = self.client.bulk_update_resources(bulk_update_items)
    except requests.exceptions.RequestException as exc:
        self._log(
            f"Bulk resource update failed with network error: {exc}. "
            f"Items will be retried on next page fetch.",
            logging.WARNING,
        )
        return 0
    # ... rest of method
```

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -253,6 +253,100 @@ def _process_and_migrate_resource_item(self, upstream_resource_item: Dict[str, A
...
+        if resp.status_code != 200:
+            self._log(
+                f"Bulk resource update returned HTTP {resp.status_code}. Items will be retried on next page fetch. Response: {resp.text[:500]}",
+                logging.WARNING,
+            )
+            return 0
+
+        resp_data = resp.json()
```
jay-steurer: **Critical: Unhandled `JSONDecodeError` on malformed response body**

If the upstream returns HTTP 200 with a non-JSON body (e.g., HTML error page from a reverse proxy), `resp.json()` raises `json.JSONDecodeError` (a subclass of `ValueError`). This crashes the migration after local resources are already committed.

**Suggested fix:**
```python
try:
    resp_data = resp.json()
except ValueError:
    self._log(
        f"Bulk resource update returned HTTP {resp.status_code} but response body "
        f"was not valid JSON. Items will be retried. Response: {resp.text[:500]}",
        logging.WARNING,
    )
    return 0
```

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -253,6 +253,100 @@ def _process_and_migrate_resource_item(self, upstream_resource_item: Dict[str, A
...
+
+            if create_gateway_resource:
+                create_operations.append((resource_type, upstream_resource["resource_data"], resource_creation_kwargs))
+
+            bulk_update_items.append(self._build_bulk_update_item(resource_ansible_id, updated_service_resource))
+
+        # Create gateway resources locally. If a resource already exists (e.g. from a
+        # previous interrupted run), the reconcile logic above will have set
+        # create_gateway_resource=False, so duplicates are safe.
+        with transaction.atomic():
```
jay-steurer, line 370: **Important: Atomicity between local creates and upstream update is intentionally broken — document it**

The old code wrapped both `Resource.create_resource()` AND `self.client.update_resource()` inside a single `transaction.atomic()`. The PR description claims "Transaction rollback on bulk-update failure (no partial state)" — but that's no longer true. Local creates commit here, and the bulk update on line 345 happens outside the transaction.

This is a reasonable trade-off for batch performance, and retries handle it via `_reconcile_existing_resource`. But the PR description should be corrected, and this docstring should document the intentional non-atomicity:

```python
# NOTE: Local creates are committed BEFORE the bulk update is sent.
# If the bulk update fails, local resources persist but upstream is not
# updated. This is safe because _reconcile_existing_resource detects
# existing resources on retry, and the upstream filter naturally retries
# unmigrated items.
```

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -253,6 +253,100 @@ def _process_and_migrate_resource_item(self, upstream_resource_item: Dict[str, A
...
+            for rt, resource_data, creation_kwargs in create_operations:
+                Resource.create_resource(rt, resource_data, **creation_kwargs)
+
+        # Send bulk update to upstream service.
+        if bulk_update_items:
+            return self._send_bulk_update(bulk_update_items)
+
+        return len(results)
```
jay-steurer: **Important: Misleading fallback — `return len(results)` implies all items succeeded**

This line is only reachable when `bulk_update_items` is empty, which only happens when `results` is empty (since every loop iteration appends to `bulk_update_items`). So it effectively returns 0, which is correct.

However, the expression `len(results)` reads as "all items were successfully processed." If a future refactor introduces a `continue` or conditional skip in the loop, this would silently over-count progress.

**Suggested fix:** `return 0` with a comment:
```python
# No items to update upstream (results was empty).
return 0
```

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -328,7 +422,6 @@ def migrate_resource(self, resource_type_name: str) -> None:
                 self._log("No more items remaining to migrate.", logging.INFO)
                 break
 
-            for upstream_resource_item in results:
-                resource_processed += 1
-                self._log_progress(progress_label, resource_processed, resource_total)
-                self._process_and_migrate_resource_item(upstream_resource_item, resource_context)
+            batch_size = self._process_resource_page_batch(results, resource_context)
```
jay-steurer, line 458: **Critical: No circuit breaker — persistent upstream failure causes an infinite loop**

When `_send_bulk_update` returns 0 (HTTP failure, network error, malformed JSON), `resource_processed` doesn't increment. The `while True` loop fetches the same page again (items reappear because `service_id` was never updated). With no failure counter, max retries, or backoff, this loops forever, hammering the upstream service with identical requests.

**Suggested fix:** Add a consecutive-failure counter:
```python
consecutive_zero_progress = 0
MAX_ZERO_PROGRESS_PAGES = 3

# ... inside the loop:
batch_size = self._process_resource_page_batch(results, resource_context)
if batch_size == 0:
    consecutive_zero_progress += 1
    if consecutive_zero_progress >= MAX_ZERO_PROGRESS_PAGES:
        raise RuntimeError(
            f"Migration stalled: {MAX_ZERO_PROGRESS_PAGES} consecutive pages "
            f"made no forward progress. Check upstream service availability."
        )
else:
    consecutive_zero_progress = 0
resource_processed += batch_size
```

jay-steurer (commented):
## PR Review Summary — [AAP-82119] Batch Resource Migration

**Files**: 3 changed (370 additions, 4 deletions)

The batch refactor is a solid performance improvement. The inline comments below flag issues that should be addressed before merge — two critical (crash scenarios after local commits), two important (atomicity semantics, dead code), and a few suggestions.

### Critical (2)
1. Unhandled `JSONDecodeError` / network exceptions in `_send_bulk_update` can crash the migration after local resources are committed
2. No circuit breaker — persistent upstream failure causes an infinite retry loop

### Important (3)
3. Local creates and upstream update are no longer atomic (contradicts PR description)
4. `_process_and_migrate_resource_item` is now dead code
5. `return len(results)` fallback has misleading semantics

### Suggestions (2)
6. All-or-nothing batch validation reduces per-item resilience vs the old code
7. Progress counter semantics changed — can confuse operators on partial failures

---
*Assisted-by: Claude Code*

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -253,6 +253,100 @@ def _process_and_migrate_resource_item(self, upstream_resource_item: Dict[str, A
...
+            resource_context: Static data about the resource type
+
+        Returns:
+            Number of items successfully processed in this batch
+        """
+        resource_type = resource_context["type"]
+        bulk_update_items = []
+        create_operations = []
+
+        for upstream_resource_item in results:
```
jay-steurer, line 339: **Suggestion: All-or-nothing batch validation reduces per-item resilience**

In the old per-item code, if item 5 of 10 failed validation, items 1-4 were fully committed. In this batch approach, any exception during the validation loop (lines 314-335) aborts the entire page — no items are created or updated.

This is acceptable if these failures indicate systemic issues (e.g., incompatible DAB version). But if per-item corruption is possible, consider wrapping individual item processing in try/except and logging + skipping bad items.

Alternatively, document this as intentional:
```python
# NOTE: If any item fails validation, the entire page is aborted.
# These failures indicate systemic issues, not per-item corruption.
```

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -328,7 +422,6 @@ def migrate_resource(self, resource_type_name: str) -> None:
                 self._log("No more items remaining to migrate.", logging.INFO)
                 break
 
-            for upstream_resource_item in results:
-                resource_processed += 1
-                self._log_progress(progress_label, resource_processed, resource_total)
-                self._process_and_migrate_resource_item(upstream_resource_item, resource_context)
+            batch_size = self._process_resource_page_batch(results, resource_context)
+            resource_processed += batch_size
```
jay-steurer, line 468: **Suggestion: Progress counter semantics changed silently**

The old code incremented `resource_processed` by 1 per item attempted. The new code increments by `batch_size` (= items successfully updated upstream). With partial errors (e.g., 50-item page, 45 succeed), failed items reappear on the next fetch, and the total can exceed `resource_total`.

Consider logging the discrepancy:
```python
if batch_size < len(results):
    self._log(
        f"Only {batch_size}/{len(results)} items updated upstream. "
        f"Remaining items will be retried.",
        logging.WARNING,
    )
```

aap_gateway_api/tests/management/commands/_migrate_service_data/test_resource_migration.py
```diff
@@ -175,6 +175,270 @@ def test_process_migrate_resource_item_raises_on_missing_resource_data():
         cmd._process_and_migrate_resource_item(resource_item, resource_context)
 
 
+@pytest.mark.django_db
```
jay-steurer, line 167: **Important: `_process_and_migrate_resource_item` is now dead code**

This PR removes the only call site for `_process_and_migrate_resource_item` (in `migrate_resource`). The method on line ~206 of the production file and this test are now testing dead code. Consider removing both in this PR or a follow-up to avoid confusion about which code path is active.

Also: the test `test_process_migrate_resource_item_raises_on_missing_resource_data` above (line 168) tests the old method. The new batch method has its own equivalent test at line 178 below.

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
Dostonbek1: Fixed in latest push. `_send_bulk_update` now wraps the `bulk_update_resources()` call in `try/except requests.exceptions.RequestException` and returns 0 on failure (items will be retried on the next page fetch).

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
Dostonbek1: Fixed. Added `try/except ValueError` around `resp.json()` — if the response body is not valid JSON, we log a warning with the first 500 chars of the response text and return 0.

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
Dostonbek1: Good catch — updated the PR description and added a NOTE comment above the `transaction.atomic()` block documenting the intentional non-atomicity:

```python
# NOTE: Local creates are committed BEFORE the bulk update is sent.
# If the bulk update fails, local resources persist but upstream is not
# updated. This is safe because _reconcile_existing_resource detects
# existing resources on retry, and the upstream filter naturally retries
# unmigrated items.
```

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
Dostonbek1: Fixed — changed to `return 0` with a clarifying comment: `# No items to update upstream (results was empty or all filtered out).`

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
Dostonbek1: Fixed — added a `consecutive_zero_progress` counter. After 3 consecutive pages with zero forward progress, the migration raises `RuntimeError("Migration stalled: ...")` instead of looping forever. The counter resets to 0 whenever a page makes progress.

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
Dostonbek1: Agreed that this is intentional — validation failures here indicate systemic issues (incompatible DAB version, corrupt upstream data) not per-item corruption. Added a NOTE comment documenting this:

```python
# NOTE: If any item fails validation, the entire page is aborted via exception.
# Validation failures indicate systemic issues (incompatible DAB version, corrupt
# data) rather than per-item corruption, so halting is the correct behavior.
```

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
Dostonbek1: Good point — added a warning log when `batch_size < len(results)`:

```python
if batch_size < len(results):
    self._log(
        f"Only {batch_size}/{len(results)} items updated upstream. Remaining items will be retried.",
        logging.WARNING,
    )
```

This makes it explicit to operators that partial failures occurred and items will be retried.

aap_gateway_api/tests/management/commands/_migrate_service_data/test_resource_migration.py
Dostonbek1: Removed both the dead `_process_and_migrate_resource_item` method and its test (`test_process_migrate_resource_item_raises_on_missing_resource_data`). The batch method and its tests now cover the equivalent behavior.

john-westcott-iv:
## Code Review

Good refactoring direction — reducing N HTTP calls per page to 1 is the right move. The error handling in `_send_bulk_update` is solid and the circuit breaker is a nice addition. Several items to address before this is ready.

### Behavior change: local creates committed before upstream confirmation

**The old code** wrapped both the local create AND the upstream HTTP call in a single `transaction.atomic()` — if upstream failed, the local create rolled back. **The new code** commits all local creates first (line 310-312), then sends the bulk update. If the bulk update fails, local resources persist with no upstream record of them.

The comment says reconciliation handles this, but it's a behavior change. To preserve the old semantics:
1. Start a transaction before local creates
2. Send the bulk update
3. For per-item failures from upstream, rollback the corresponding local resources
4. Only commit items that succeeded both locally and upstream

### Items not chunked to match upstream limit

The DAB bulk endpoint enforces `MAX_BULK_SIZE = 1000`, but this PR doesn't chunk items before sending. If a page returns > 1000 items, the bulk call gets a 400. Either chunk `bulk_update_items` into batches of 1000, or document/enforce that page size must be <= 1000.

### `_send_bulk_update` return type undocumented

The docstring says "return the number of successfully updated items" but the method signature lacks `-> int` and the Returns section is missing. Minor, but readers shouldn't have to read the implementation to know the return type.

### `import requests` just for the exception (line 228)

The migration code shouldn't need to know about the HTTP library. Either:
- Have `bulk_update_resources` catch `requests.exceptions.RequestException` internally and raise a domain-specific exception
- Or use a generic `except Exception` here since `_send_bulk_update` is already non-fatal by design

### Non-200 responses need transient vs permanent distinction (line 235)

All non-200 responses are treated the same: log, return 0, retry next page. But:
- **502/503/504** — transient, retry immediately with backoff
- **400** — permanent, payload is invalid, retrying the same data is pointless
- **401/403** — auth issue, retrying won't help
- **413** — payload too large, need to chunk

The "items will be retried on next page fetch" message is misleading for 400 errors — the same payload will produce the same 400.

### Invalid JSON response (line 242)

Same issue — a 200 with non-JSON body is likely a proxy error (nginx HTML 502). Transient, worth distinguishing from a permanent failure.

### `raise RuntimeError` on missing `resource_data` (line 288)

This aborts the ENTIRE migration, not just this page. In the old per-item code, one bad item only killed that item. In batch context, one item missing `resource_data` kills everything — including items that were fine. The comment says "systemic issues," but what if one service is on an older DAB version? Consider making this a per-item error instead.

### Newly created items also queued for bulk update (lines 302-305)

I understand the upstream update is telling the other service "I've imported this, update your records" — it's not updating the same resource we just created. But this should be atomic with the local create (per the behavior preservation point above). If the upstream update fails for this item, the local create should roll back.

### Circuit breaker is too blunt (lines 386-387)

3 consecutive zero-progress pages → die. This doesn't distinguish between transient failures (worth retrying) and permanent ones (bad data that will never succeed). Better approach:
1. On transient HTTP error (502/503/timeout): retry the bulk call up to 3 times with backoff
2. On permanent per-item errors (400 from bad data): remove the failing items, resend only the good ones
3. If zero good items can pass after isolation: log the permanently-failing items, move to next page
4. Only abort if ALL pages across a full cycle make zero progress

This way transient errors get retried, permanent per-item failures get isolated, and the migration only aborts when truly stuck.

### "Remaining items will be retried" (line 416) — when?

The log message says items will be retried, but the code moves to the next page fetch. They'll reappear on the next fetch because they're still unmigrated, but if the failure is permanent (bad data), they'll reappear on EVERY fetch and eventually trigger the circuit breaker. This is where isolating permanently-failing items (point above) would help — let the good items through instead of blocking the whole page.

### CI failure

`test_migrate_resource_circuit_breaker` fails with `AttributeError: 'Command' object has no attribute '_progress_thresholds'` — mock setup is missing required attributes.

Dostonbek1:
Thanks for the detailed review @john-westcott-iv! Here's how each point was addressed:

### 1. Behavior change: local creates before upstream confirmation

Intentional design choice — keeping current behavior. Wrapping local DB creates in a transaction that spans an HTTP call is an anti-pattern (holds DB locks for network I/O duration, risk of connection pool exhaustion under load). The current approach is safe because:
- `_reconcile_existing_resource` detects existing resources on retry (idempotent)
- Failed upstream updates leave items unmigrated, so the filter naturally retries them on next iteration
- This matches the pattern used by role assignment migration in the same codebase

Added inline comment documenting this design decision.

### 2. Items not chunked to match upstream limit ✅ Fixed

Added `MAX_BULK_CHUNK_SIZE = 1000`. `_send_bulk_update` now chunks items and sums the results. Added test (`test_send_bulk_update_chunks_large_batches`) verifying 2500 items are sent in 3 chunks.

### 3. `_send_bulk_update` return type undocumented ✅ Fixed

Added `-> int` type annotation and expanded the docstring.

### 4. `import requests` just for the exception

Keeping the specific `requests.exceptions.RequestException` catch. The `rest_client` already depends on `requests` internally — we're not introducing a new dependency. Being specific about which exception type we expect is better for debugging than a bare `except Exception`.

### 5/6. Non-200 responses & invalid JSON need transient vs permanent distinction ✅ Fixed

Completely reworked `_send_bulk_update` into a two-layer design:
- `_send_bulk_update` — chunks items into batches of 1000
- `_send_bulk_update_chunk` — sends a single chunk with retry logic:
  - **502/503/504 + network errors**: Retry up to 3 times with exponential backoff (1s, 2s, 4s)
  - **4xx (permanent)**: Fail immediately, no retry, clear log message
  - **Non-JSON response**: Treated as transient (likely proxy error), retried

Added tests: `test_send_bulk_update_permanent_error`, `test_send_bulk_update_transient_then_success`.

### 7. `raise RuntimeError` on missing `resource_data`

Keeping the hard abort. This isn't a per-item issue — if one resource lacks `resource_data`, ALL resources from that service will too (it's a DAB version/feature-flag thing, not per-item corruption). The abort draws immediate attention to the version mismatch.

### 8. Newly created items queued for bulk update — atomicity

Same rationale as #1. The bulk update is telling the upstream service "I've imported these, update your records." It's not modifying the same resources we just created locally. If it fails, the local resources exist but upstream isn't updated — next iteration detects them via `_reconcile_existing_resource` and skips re-creation.

### 9. Circuit breaker is too blunt ✅ Partially addressed

The circuit breaker now benefits from the retry logic in `_send_bulk_update_chunk` — transient errors (502, network) are retried 3 times with backoff before returning 0. So the circuit breaker only triggers after 3 consecutive pages where ALL retry attempts were exhausted. This is a reasonable threshold for "truly stuck."

Full per-item failure isolation (removing permanently-failing items and resending) is a larger refactor that would require the upstream to distinguish "this item's data is permanently invalid" from "transient issue" — currently all per-item errors look the same in the response. Worth exploring in a follow-up.

### 10. "Remaining items will be retried" — when? ✅ Fixed

Updated log message to: "Failed items remain unmigrated and will reappear on the next page fetch." This makes the mechanism clear.

### 11. CI failure ✅ Fixed (previous push)

Was missing `cmd._progress_thresholds = {}` in the test setup. Fixed in an earlier commit.

---

All changes pushed in the latest commit.

jay-steurer (approved):

PR #171, opened 2026-07-22
[AAP-82668] Wrap org delete with deferral context managers

AlanCoding:
## Summary

- Wraps `OrganizationViewSet.destroy` with `defer_rbac_computations`, `deferred_activity_stream`, and `defer_resource_cleanup` context managers so cascade deletes of large organizations batch all signal-driven work instead of firing per-object.
- Addresses organization deletion timeout at scale (AAP-82668).

Requires ansible/django-ansible-base#1059

## Test plan

- [ ] Delete an organization with many teams/users/inventories — should complete without timeout
- [ ] Delete a managed organization — should still return 400
- [ ] Delete a small organization — should behave identically to before

### DIFF
--- a/aap_gateway_api/views/api/v1/organization.py
+++ b/aap_gateway_api/views/api/v1/organization.py
@@ -1,5 +1,8 @@
 import logging
 
+from ansible_base.activitystream import deferred_activity_stream
+from ansible_base.rbac.triggers import defer_rbac_computations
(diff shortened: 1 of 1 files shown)


AlanCoding:
Getting some thoughts that people would rather this be done in the base view in DAB. That might be possible and I'll look into it.

AlanCoding:
Closing because the aforementioned tweak has been included in the main patch.

PR #174, opened 2026-07-23
[AAP-83975] Scope ext_auth AUTH_TYPE_NONE to EDA routes only

Dostonbek1:
## Summary

When `enable_gateway_auth=False`, the ext_auth filter now uses `AUTH_TYPE_NONE` only for EDA routes (to add `X-Trusted-Proxy` for webhook verification). All other services completely disable ext_auth (`disabled: True`), restoring the original behavior.

Requires ansible/ansible.platform#219

## Changes

- `aap_gateway_api/common/envoy.py` — Added `AUTH_TYPE_NONE` constant
- `aap_gateway_api/models/route.py` — Added `is_eda_service` property; EDA routes get `AUTH_TYPE_NONE`, all others get `disabled: True`
- `aap_gateway_api/models/service_type.py` — Added `is_eda_service` cached property
- `aap_gateway_api/proxy/control_plane.py` — Handle `AUTH_TYPE_NONE` by adding `X-Trusted-Proxy` without authentication
- Tests updated across `test_route.py`, `test_proxy.py`, and `test_service_cluster.py`

## Why

The original CVE fix applied `AUTH_TYPE_NONE` to all `enable_gateway_auth=False` routes, which caused two issues:
1. The `X-Trusted-Proxy` header triggered `DisallowWriteFromProxy` on gateway config endpoints (403 on PATCH)
2. Unnecessary gRPC calls to the control plane were misrouted through corporate proxies in restricted network environments (401 JWT errors)

Scoping `AUTH_TYPE_NONE` to EDA only eliminates both issues while preserving the CVE fix for the one service that needs it.

## Test plan
- [ ] CI pipeline green
- [ ] Unit tests pass (ruff-check, ruff-format clean)

---

### DIFF
--- a/aap_gateway_api/models/route.py
+++ b/aap_gateway_api/models/route.py
@@ -124,9 +124,9 @@ class Route(UniqueNamedCommonModel, AuditableModel):
     )
 
     @property
-    def is_gateway_service(self):
-        """True if this route targets the gateway service itself."""
-        return self.service_cluster.service_type.is_gateway_service
+    def is_eda_service(self):
+        """True if this route targets the EDA service."""
+        return self.service_cluster.service_type.is_eda_service
 
     def is_internal_route_string(self):
         return "t" if self.is_internal_route else "f"
@@ -286,12 +286,9 @@ class Route(UniqueNamedCommonModel, AuditableModel):
             }
 
         if not self.enable_gateway_auth:
-            if self.is_gateway_service:
-                cfg["typed_per_filter_config"][EXT_AUTH_FILTER] = {
-                    TYPE_KEY: EXT_AUTH_PER_ROUTE,
-                    "disabled": True,
-                }
-            else:
+            if self.is_eda_service:
+                # EDA webhooks handle their own auth but need X-Trusted-Proxy
+                # to verify requests originated from the gateway (CVE fix)
                 cfg["typed_per_filter_config"][EXT_AUTH_FILTER] = {
                     TYPE_KEY: EXT_AUTH_PER_ROUTE,
                     "check_settings": {
@@ -301,6 +298,15 @@ class Route(UniqueNamedCommonModel, AuditableModel):
                             "auth_type": AUTH_TYPE_NONE,
                         },
                     },
+                }
+            else:
+                # All other services: completely disable ext_auth to restore
+                # pre-CVE-fix behavior. This avoids X-Trusted-Proxy being added
+                # (which causes 403 on gateway config writes) and prevents
+                # unnecessary gRPC calls (which causes 401 in proxy environments).
+                cfg["typed_per_filter_config"][EXT_AUTH_FILTER] = {
+                    TYPE_KEY: EXT_AUTH_PER_ROUTE,
+                    "disabled": True,
                 }
         else:
             cfg["typed_per_filter_config"][EXT_AUTH_FILTER] = {
--- a/aap_gateway_api/models/route.py
+++ b/aap_gateway_api/models/route.py
@@ -6,7 +6,7 @@ from django.db import models
 from django.urls import reverse
 from django.utils.translation import gettext as _
 
-from aap_gateway_api.common.envoy import EXT_AUTH_FILTER, EXT_AUTH_PER_ROUTE, LUA_PER_ROUTE, UPSTREAM_TLS_CONTEXT
+from aap_gateway_api.common.envoy import AUTH_TYPE_NONE, EXT_AUTH_FILTER, EXT_AUTH_PER_ROUTE, LUA_PER_ROUTE, UPSTREAM_TLS_CONTEXT
 from aap_gateway_api.models.http_port import HTTPPort
 from aap_gateway_api.models.service_cluster import ServiceCluster
 from aap_gateway_api.models.service_type import DefaultServiceType
@@ -283,7 +283,13 @@ class Route(UniqueNamedCommonModel, AuditableModel):
         if not self.enable_gateway_auth:
             cfg["typed_per_filter_config"][EXT_AUTH_FILTER] = {
                 TYPE_KEY: EXT_AUTH_PER_ROUTE,
-                "disabled": not self.enable_gateway_auth,
+                "check_settings": {
+                    "context_extensions": {
+                        "is_internal_route": self.is_internal_route_string(),
+                        "service_type": self.service_cluster.service_type.name,
+                        "auth_type": AUTH_TYPE_NONE,
+                    },
+                },
             }
         else:
             cfg["typed_per_filter_config"][EXT_AUTH_FILTER] = {
--- a/aap_gateway_api/models/route.py
+++ b/aap_gateway_api/models/route.py
@@ -123,6 +123,11 @@ class Route(UniqueNamedCommonModel, AuditableModel):
         ),
     )
 
+    @property
+    def is_gateway_service(self):
+        """True if this route targets the gateway service itself."""
+        return self.service_cluster.service_type.is_gateway_service
+
     def is_internal_route_string(self):
         return "t" if self.is_internal_route else "f"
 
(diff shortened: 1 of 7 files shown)


aap_gateway_api/proxy/control_plane.py
```diff
@@ -267,6 +268,7 @@ def Check(self, request, context):
 
         self.headers = []
         self.service_type = request.attributes.context_extensions["service_type"]
+        self.auth_type = request.attributes.context_extensions["auth_type"]
```
jay-steurer, line 271: **Critical: Unguarded `KeyError` on `auth_type` dict access**

This uses bracket-style dict access outside any try-except. If `auth_type` is ever missing from `context_extensions` (rolling deploy with config/binary out of sync, misconfigured filter chain, future route type), this crashes the gRPC handler with a bare `KeyError` and no actionable context.

Depending on Envoy's `failure_mode_allow` setting, this either **blocks all traffic** to the affected route or **allows unauthenticated passthrough**.

Note: `service_type` on the line above has the same pre-existing pattern, but `should_reject_failed_basic_auth` already uses `.get()` — the codebase is inconsistent.

Suggested fix:
```python
self.auth_type = request.attributes.context_extensions.get("auth_type")
if self.auth_type is None:
    logger.error(
        "Missing auth_type in context_extensions for %s (service_type=%s). "
        "This indicates an xDS route misconfiguration.",
        request.attributes.request.http.path, self.service_type,
    )
    raise ValueError(f"Missing required auth_type in context_extensions for {self.service_type}")
```

aap_gateway_api/proxy/control_plane.py
```diff
@@ -295,6 +297,13 @@ def Check(self, request, context):
         self.is_internal_route = self.is_route_internal(request)
         self.reject_failed_basic_auth = self.should_reject_failed_basic_auth(request)
 
+        # Routes with auth_type=NONE (enable_gateway_auth=false) skip authentication
+        # but still get the X-Trusted-Proxy header to verify they came through the gateway.
+        # This is useful for services like EDA that handle their own authentication.
+        if self.auth_type == AUTH_TYPE_NONE:
+            logger.debug(f"Auth type is NONE for {self.request_id} {self.request_path} - adding trusted proxy header without authentication")
+            return self._return_no_authentication_required()
```
jay-steurer, line 305: **Important: `_add_trusted_proxy_header` silently swallows exceptions on this path** (pre-existing, but now load-bearing)

The entire purpose of `AUTH_TYPE_NONE` is to add `X-Trusted-Proxy` for EDA webhooks. `_return_no_authentication_required()` calls `_add_trusted_proxy_header()`, which catches *all* exceptions:
```python
except Exception:
    logger.exception("Failed to generate x-trusted-proxy")
```
If `get_jwt_rsa_key()` or JWT signing fails, the response is returned **without** the header — silently disabling the CVE protection. For `AUTH_TYPE_NONE` routes, a failure to add `X-Trusted-Proxy` should be a hard failure (fail closed), not swallowed.

**Suggestion:** Consider either:
- Letting the exception propagate for `AUTH_TYPE_NONE` routes (e.g., `_return_no_authentication_required(require_trusted_proxy=True)`)
- Or at minimum, escalating the log to include `SECURITY:` so it's monitorable

aap_gateway_api/proxy/control_plane.py
```diff
@@ -295,6 +297,13 @@ def Check(self, request, context):
         self.is_internal_route = self.is_route_internal(request)
         self.reject_failed_basic_auth = self.should_reject_failed_basic_auth(request)
 
+        # Routes with auth_type=NONE (enable_gateway_auth=false) skip authentication
+        # but still get the X-Trusted-Proxy header to verify they came through the gateway.
+        # This is useful for services like EDA that handle their own authentication.
```
jay-steurer, line 302: **Nit:** "services like EDA" implies other services use this pattern, but only EDA routes can reach this code path (gated by `is_eda_service` in `route.py`). Consider:
```python
# Used by EDA, which handles its own webhook authentication.
```

aap_gateway_api/proxy/control_plane.py
```diff
@@ -295,6 +297,13 @@ def Check(self, request, context):
...
+        if self.auth_type == AUTH_TYPE_NONE:
```
jay-steurer, line 303: **Suggestion: Move this check earlier in `Check()`**

For `AUTH_TYPE_NONE` routes, `get_drf_request()` (Django request construction, session middleware, auth classes) and `_ensure_db_availability()` all run before this short-circuit. None of that work is needed — and a middleware failure could block a webhook that requires no Django processing.

Moving this check right after the `auth_type` extraction would be both faster and more resilient.

aap_gateway_api/models/service_type.py
```diff
@@ -13,6 +14,11 @@ class ServiceType(UniqueNamedCommonModel, AuditableModel):
 
     router_basename = 'service_type'
 
+    @cached_property
+    def is_eda_service(self):
+        """True if this is the EDA service type."""
+        return self.name == DefaultServiceType.EDA
```
jay-steurer, line 20: **Important: Use `DefaultServiceType.EDA.value` instead of `DefaultServiceType.EDA`**

This comparison works because `DefaultServiceType` inherits from `str`. If the enum base class is ever changed to plain `Enum`, this would silently return `False` for *all* service types — disabling `AUTH_TYPE_NONE` for EDA with no error raised.

The existing `is_default()` method in this same file explicitly uses `.value`:
```python
return any(svc.value == name for svc in DefaultServiceType)
```

Suggested fix:
```python
return self.name == DefaultServiceType.EDA.value
```

aap_gateway_api/models/service_type.py
```diff
@@ -13,6 +14,11 @@ class ServiceType(UniqueNamedCommonModel, AuditableModel):
 
     router_basename = 'service_type'
 
+    @cached_property
```
jay-steurer, line 17: **Suggestion: Use `@property` instead of `@cached_property`**

`cached_property` stores the result in the instance's `__dict__`. Django's `refresh_from_db()` updates field values but does *not* clear `cached_property` entries. If a `ServiceType` name were ever mutated and the same instance reused, the cached value would be stale.

The performance gain of caching a single string comparison is negligible. A plain `@property` eliminates the stale-state risk:
```python
@property
def is_eda_service(self):
    return self.name == DefaultServiceType.EDA.value
```

aap_gateway_api/models/route.py
```diff
@@ -281,10 +286,28 @@ def get_xds_route_config(self):
             }
 
         if not self.enable_gateway_auth:
-            cfg["typed_per_filter_config"][EXT_AUTH_FILTER] = {
-                TYPE_KEY: EXT_AUTH_PER_ROUTE,
-                "disabled": not self.enable_gateway_auth,
-            }
+            if self.is_eda_service:
+                # EDA webhooks handle their own auth but need X-Trusted-Proxy
+                # to verify requests originated from the gateway (CVE fix)
```
jay-steurer, line 291: **Nit:** "(CVE fix)" without a ticket number — the test docstrings reference AAP-79215/79216 but this production comment doesn't. A future reader can't trace it without git blame. Suggest:
```python
# so EDA can verify requests originated from the gateway (AAP-79215).
```

aap_gateway_api/models/route.py
```diff
@@ -281,10 +286,28 @@ def get_xds_route_config(self):
...
+                    "check_settings": {
+                        "context_extensions": {
+                            "is_internal_route": self.is_internal_route_string(),
+                            "service_type": self.service_cluster.service_type.name,
+                            "auth_type": AUTH_TYPE_NONE,
+                        },
+                    },
+                }
+            else:
+                # All other services: completely disable ext_auth to restore
```
jay-steurer, line 303: **Nit:** "pre-CVE-fix behavior" is a time-relative reference that will rot. The parenthetical explanations are excellent though. Consider removing just the temporal clause:
```python
# All other services: completely disable ext_auth. This avoids
# X-Trusted-Proxy being added (which triggers DisallowWriteFromProxy
# 403s on gateway config writes) and prevents unnecessary gRPC calls
# to the control plane (which fail with 401 in proxy environments).
```
Linking the 403 to `DisallowWriteFromProxy` also makes it grep-able.

jay-steurer (commented):
## PR Review Summary

**Overall:** Well-implemented, security model is sound. The scoping of `AUTH_TYPE_NONE` to EDA-only routes is the correct fix — it preserves the CVE mitigation while eliminating side effects on other services.

**Strengths:**
- `context_extensions` are server-controlled via xDS — not client-spoofable
- `AUTH_TYPE_NONE` sentinel can't collide with valid `ServiceAuthType` choices (JWT/BASIC/TOKEN)
- Critical invariant tested from both directions (EDA gets NONE, non-EDA gets `disabled: True`)
- End-to-end regression covered in `test_service_cluster.py`
- CI is green (single failure is infra-related Node.js 20 deprecation)

See inline comments for specific findings. 1 critical, 2 important, and several suggestions.

jay-steurer (approved):

Dostonbek1 (commented):
## Forward-port Review

**Diff Comparison**: Expected differences only (branch divergence from downstream repo)
**CI Status**: All 24 checks passing

### Differences Categorized

| Category | Difference | Verdict |
|----------|-----------|---------|
| New file changes | `envoy.py` — devel uses protobuf-derived constants, `AUTH_TYPE_NONE` added alongside them | Expected |
| New file changes | `control_plane.py` — `AUTH_TYPE_NONE` handling added (import, `self.auth_type` init, NONE check) | Expected |
| New file changes | `test_proxy.py` — `TestAuthTypeNone` class added; devel keeps its own profiling test classes | Expected |
| Import changes | `route.py` — devel import includes `LUA_PER_ROUTE, UPSTREAM_TLS_CONTEXT` | Branch-specific |
| Property addition | `service_type.py` — `from functools import cached_property` added (downstream already had it) | Expected |
| Context lines | `route.py` — devel has `reject_failed_basic_auth` in `context_extensions` | Branch-specific |
| Test assertion style | `test_service_cluster.py` — minor assertion order fix | Acceptable |

### Core Logic Comparison

The core logic is **identical** across both branches:
- `is_eda_service` property on Route and ServiceType
- EDA routes get `AUTH_TYPE_NONE`, all others get `disabled: True`
- Same test structure for route config, EDA event stream, and non-EDA services

**Review Summary**:
- Forward-port accurately reflects the downstream changes
- All differences are expected branch divergence, not logic changes
- All CI checks are passing

**Verdict**: LGTM — ready for review from another maintainer.

PR #179, opened 2026-07-28
[AAP-62944] Ensure health check interval >= effective timeout

bhavenst:
## Summary

Under peak API load, Envoy health checks can time out and falsely eject healthy upstream nodes, causing 503 errors. The health check interval (default 10s) could be shorter than the effective health check timeout (at least 30s), causing overlapping health checks and potentially invalid Envoy configuration.

This PR adds `get_effective_health_check_interval_seconds()` to `ServiceCluster` which floors the interval at the effective timeout, mirroring the existing `get_effective_health_check_timeout_seconds()` pattern. It uses the effective interval in xDS cluster config generation, exposes it as a read-only `effective_health_check_interval_seconds` field on the API, and raises the model default from 10s to 30s for new clusters. No data migration — existing clusters keep their stored values and the effective-interval logic handles correctness at runtime.

## Acceptance Criteria

From the [Jira ticket](https://redhat.atlassian.net/browse/AAP-62944) Expected Behavior:

| # | Criterion | Status | Where |
|---|-----------|--------|-------|
| 1 | **Health check timeouts should align with `request_timeout`** — nodes should only be ejected if truly unresponsive, not just busy. The default should align with whatever `request_timeout` the operator/installer configured. | Already done | Addressed by [AAP-66486](https://redhat.atlassian.net/browse/AAP-66486) (commit `e2729e30`), which added `get_effective_health_check_timeout_seconds()` = `max(cluster_hc_timeout, max_route_timeout, global_request_timeout)`. Effective timeout is at least 30s by default. |
| 2 | **Health check frequency must decrease to match raised timeouts** — e.g., if timeout is 15s, interval should be no less than the timeout to avoid overlapping checks. | This PR | Adds `get_effective_health_check_interval_seconds()` = `max(stored_interval, effective_timeout)`. Uses it in `get_xds_cluster_config()` so Envoy always receives `interval >= timeout`. Raises model default from 10s to 30s for new clusters. |
| 3 | **Values must remain configurable via the API** — admins can tune health check parameters for their specific workload profiles. | Already done + this PR | All health check fields (`health_check_timeout_seconds`, `health_check_interval_seconds`, `health_check_unhealthy_threshold`, `health_check_healthy_threshold`) remain writable on the ServiceCluster API. This PR adds a read-only `effective_health_check_interval_seconds` computed field (matching the existing `effective_health_check_timeout_seconds`) so admins can see the actual value Envoy will use. |
| 4 | **Gateway should allow increased latency during peak loads without cutting off service access** — no false-negative outages from busy-but-healthy nodes. | Combination | Criteria 1-3 together ensure that under peak load, health checks have sufficient timeout to complete and sufficient interval to avoid overlapping, while still allowing Envoy to detect truly unresponsive nodes. |

## Test plan
- [x] Parametrized model tests for `get_effective_health_check_interval_seconds()` (interval below/above effective timeout, with route timeouts)
- [x] Integration test verifying `get_xds_cluster_config()` uses the effective interval
- [x] Serializer tests for the new `effective_health_check_interval_seconds` computed field
- [x] CDS view integration test verifying Envoy receives the correct interval
- [x] All 33 existing `health_check` tests pass
- [x] Migration check passes (`tox -e check-migrations`)
- [x] Cross-checked live Gateway API `effective_*` fields against Envoy admin config dump — values match

Closes: https://redhat.atlassian.net/browse/AAP-62944

[AAP-66486]: https://redhat.atlassian.net/browse/AAP-66486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Summary by CodeRabbit

* **New Features**
  * Service cluster API responses now include the computed *effective health-check interval*.
* **Improvements**
  * Default service cluster health-check interval is now **30 seconds**.
  * Generated xDS health-check configuration now uses the effective interval, derived from effective timeout and request-timeout preferences (including proxy settings).
* **Bug Fixes**
  * Health-check `interval` and `timeout` are now consistently aligned, ensuring the interval is never shorter than the effective timeout.
* **Tests**
  * Added coverage for interval “flooring” behavior and serializer/xDS propagation.

### DIFF
--- a/aap_gateway_api/models/route.py
+++ b/aap_gateway_api/models/route.py
@@ -205,10 +205,11 @@ class Route(UniqueNamedCommonModel, AuditableModel):
 
         if self.service_cluster.health_checks_enabled:
             effective_health_check_timeout = self.service_cluster.get_effective_health_check_timeout_seconds()
+            effective_health_check_interval = self.service_cluster.get_effective_health_check_interval_seconds()
             cfg["health_checks"] = [
                 {
                     "timeout": f"{effective_health_check_timeout}s",
-                    "interval": f"{self.service_cluster.health_check_interval_seconds}s",
+                    "interval": f"{effective_health_check_interval}s",
                     "unhealthy_threshold": self.service_cluster.health_check_unhealthy_threshold,
                     "healthy_threshold": self.service_cluster.health_check_healthy_threshold,
                     "http_health_check": {
--- a/aap_gateway_api/models/route.py
+++ b/aap_gateway_api/models/route.py
@@ -205,7 +205,7 @@ class Route(UniqueNamedCommonModel, AuditableModel):
 
         if self.service_cluster.health_checks_enabled:
             effective_health_check_timeout = self.service_cluster.get_effective_health_check_timeout_seconds()
-            effective_health_check_interval = self.service_cluster.get_effective_health_check_interval_seconds()
+            effective_health_check_interval = self.service_cluster.get_effective_health_check_interval_seconds(effective_health_check_timeout)
             cfg["health_checks"] = [
                 {
                     "timeout": f"{effective_health_check_timeout}s",
--- a/aap_gateway_api/models/service_cluster.py
+++ b/aap_gateway_api/models/service_cluster.py
@@ -66,7 +66,7 @@ class ServiceCluster(UniqueNamedCommonModel, AuditableModel):
     )
 
     health_check_interval_seconds = models.PositiveIntegerField(
-        default=10,
+        default=30,
         help_text=_("The time between health check requests."),
     )
 
@@ -203,6 +203,10 @@ class ServiceCluster(UniqueNamedCommonModel, AuditableModel):
             max_route_timeout,
             get_preference_value('proxy', 'request_timeout'),
         )
+
+    def get_effective_health_check_interval_seconds(self):
+        effective_timeout = self.get_effective_health_check_timeout_seconds()
+        return max(self.health_check_interval_seconds, effective_timeout)
 
     @staticmethod
     def get_cluster_by_type(service_type: ServiceType | str):
--- a/aap_gateway_api/models/service_cluster.py
+++ b/aap_gateway_api/models/service_cluster.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import datetime
 from typing import Optional
 
@@ -11,6 +12,8 @@ from django.utils.translation import gettext as _
 
 from aap_gateway_api.models.service_type import DefaultServiceType, ServiceType
 from aap_gateway_api.utils.preferences import get_preference_value
+
+logger = logging.getLogger(__name__)
 
 
 class ServiceCluster(UniqueNamedCommonModel, AuditableModel):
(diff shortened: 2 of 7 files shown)


aap_gateway_api/models/route.py
```diff
@@ -205,10 +205,11 @@ def get_xds_cluster_config(self):
 
         if self.service_cluster.health_checks_enabled:
             effective_health_check_timeout = self.service_cluster.get_effective_health_check_timeout_seconds()
+            effective_health_check_interval = self.service_cluster.get_effective_health_check_interval_seconds()
             cfg["health_checks"] = [
                 {
                     "timeout": f"{effective_health_check_timeout}s",
-                    "interval": f"{self.service_cluster.health_check_interval_seconds}s",
+                    "interval": f"{effective_health_check_interval}s",
```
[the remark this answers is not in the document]
bhavenst: Done Mr. Rabbit

aap_gateway_api/models/service_cluster.py
```diff
@@ -204,6 +204,11 @@ def get_effective_health_check_timeout_seconds(self):
             get_preference_value('proxy', 'request_timeout'),
         )
 
+    def get_effective_health_check_interval_seconds(self, effective_timeout=None):
+        if effective_timeout is None:
+            effective_timeout = self.get_effective_health_check_timeout_seconds()
+        return max(self.health_check_interval_seconds, effective_timeout)
+
     @staticmethod
```
jay-steurer, line 223: suggestion (non-blocking): Consider adding a `logger.debug()` when the floor logic overrides the stored interval, so operators can trace why the effective value differs from what they configured. The existing `get_effective_health_check_timeout_seconds()` also lacks this, so it would be a pattern-level improvement across both methods.

```suggestion
    def get_effective_health_check_interval_seconds(self, effective_timeout=None):
        if effective_timeout is None:
            effective_timeout = self.get_effective_health_check_timeout_seconds()
        effective_interval = max(self.health_check_interval_seconds, effective_timeout)
        if effective_interval != self.health_check_interval_seconds:
            logger.debug(
                "Health check interval for cluster %s floored from %ds to %ds (effective timeout)",
                self.name, self.health_check_interval_seconds, effective_interval,
            )
        return effective_interval
```

aap_gateway_api/tests/models/test_route.py
```diff
@@ -783,3 +783,82 @@ def test_xds_cluster_config_uses_effective_health_check_timeout(self, service_cl
...
+
+    @pytest.mark.parametrize(
+        "health_check_interval,health_check_timeout,route_timeout,preference_timeout,expected_interval",
+        [
+            (10, 5, None, 30, 30),
+            (60, 5, None, 30, 60),
+            (10, 5, 600, 30, 600),
+            (700, 5, 600, 30, 700),
+        ],
+        ids=[
```
jay-steurer, line 797: suggestion (non-blocking): Consider adding a zero-interval edge case and an equality boundary case to the parametrized set. `PositiveIntegerField` permits 0, and while `max()` handles both trivially, these serve as living documentation:

```python
(0, 5, None, 30, 30),   # zero_interval_uses_effective_timeout
(30, 5, None, 30, 30),  # interval_equals_effective_timeout
```

jay-steurer (commented):
Clean, well-tested PR that follows established patterns. Three minor suggestions below — none block merge.

bhavenst:
Seems ATF client needs re-gen to fix those failing tests.

jay-steurer (approved):

jay-steurer:
Test failures are related to another change in eda, I believe.

PR #180, opened 2026-07-29
[AAP-77766] Publish open source Jewel container image to GHCR

Dostonbek1:
## Description

- **What is being changed?** Adds a GitHub Actions workflow that builds and publishes a Jewel container image to GHCR (`ghcr.io/ansible/jewel`) on every push to `devel`. Also makes the platform UI inclusion optional in the Dockerfile via a build arg, and documents image usage in the README.
- **Why is this change needed?** Jewel currently has no published builds, making it difficult for open source contributors to use the project without building from scratch. AWX and EDA already publish upstream images — Jewel should follow suit.
- **How does this change address the issue?** Introduces a CI workflow modeled after EDA Server and AWX patterns, using GHCR with the built-in `GITHUB_TOKEN` (no custom secrets needed).

## Type of Change

- [x] Documentation update
- [x] Configuration change

## Self-Review Checklist

### Prerequisites
- Push access to a fork of `ansible/jewel`

### Steps to Test
1. Merge to `devel` — the workflow runs automatically on push
2. Check the Actions tab for the "Build and Publish Image" workflow
3. Verify the image is available at `ghcr.io/ansible/jewel:latest`
4. Pull and run: `docker pull ghcr.io/ansible/jewel:latest`

### Expected Results
- Workflow completes successfully and pushes multi-arch (amd64/arm64) image
- Image is tagged with `latest`, `sha-<short>`, and `sha-<long>`
- Image does NOT include the platform UI by default
- Building with `--build-arg INCLUDE_UI=true` includes the UI

### Design Decisions
- **GHCR over Quay.io** — aligns with AWX's approach and signals upstream/devel. No custom robot account secrets needed.
- **`latest` tag** — rolling tag updated on every push to `devel`, plus immutable SHA-based tags for pinning.
- **`INCLUDE_UI=false` by default** — per acceptance criteria, the UI reference is removed by default with a build arg to opt in.
- **Fork guard** — `if: github.repository == 'ansible/jewel'` prevents the workflow from running on forks.

### Required Actions
  - Community team should be informed once the image is live
  - GHCR package visibility may need to be set to "public" by an org admin after the first publish

---

## Summary by CodeRabbit

- **New Features**
  - Added an automated “Build and Publish Image” workflow on the `devel` branch.
  - Publishes multi-architecture container images (AMD64 and ARM64) to GHCR with `latest` and commit-based tags (`sha-<short>` / `sha-<full>`).
  - Introduced image build targets to optionally include the platform UI (default builds use a placeholder).

- **Documentation**
  - Added a “Container Image” section with pull instructions, available tag formats, and local build guidance (including optional UI inclusion).

### DIFF
--- a/.github/workflows/build-image.yml
+++ b/.github/workflows/build-image.yml
@@ -1,0 +1,64 @@
+name: Build and Publish Image
+
+on:
+  push:
+    branches:
+      - devel
+
+concurrency:
+  group: publish-image-${{ github.ref }}
+  cancel-in-progress: false
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ansible/jewel
+
+jobs:
+  build-and-push-image:
+    if: github.repository == 'ansible/jewel'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+        with:
+          persist-credentials: false
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # QEMU enables emulation for cross-platform builds (arm64 on amd64 runners)
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@c7c53464625b32c7a7e944ae62b3e17d2b600130 # v3.7.0
+
+      # Buildx is required for multi-arch image builds and direct registry push
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@8d2750c68a42422c14e847fe6c8ac0403b4cbd6f # v3.12.0
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@c299e40c65443455700f0fdfc63efafe5b349051 # v5.10.0
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=latest
+            type=sha
+            type=sha,format=long
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@10e90e3645eae34f1e60eeb005ba3a3d33f178e8 # v6.19.2
+        with:
+          context: .
+          file: tools/docker/Dockerfile
+          target: jewel
+          platforms: linux/amd64,linux/arm64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
+          labels: ${{ steps.meta.outputs.labels }}
--- a/README.md
+++ b/README.md
@@ -17,6 +17,29 @@ Jewel provides the source code for the Gateway that connects Ansible services an
 > The project name, Jewel, refers to a precious stone known as the Eye of the Sea from the same novel that originated the word Ansible, Ursula K. Le Guin's *Rocannon's World*.
 > Ansible communicates across distance; Jewel is what everything converges on.
 
+## Container Image
+
+A pre-built container image is published to the GitHub Container Registry on every push to the `devel` branch. The published image does **not** include the platform UI.
+
+```bash
+docker pull ghcr.io/ansible/jewel:latest
+docker run -p 8000:8000 ghcr.io/ansible/jewel:latest
+```
+
+Available tags:
+
+| Tag | Description |
+|---|---|
+| `latest` | Most recent build from `devel` |
+| `sha-<short>` | Pinned to a specific commit (short SHA) |
+| `sha-<full>` | Pinned to a specific commit (full SHA) |
+
+To build locally with the platform UI included:
+
+```bash
+docker build --target jewel-ui -f tools/docker/Dockerfile .
+```
+
 ## Communication
 
 Join the Ansible forum:
--- a/tools/docker/Dockerfile
+++ b/tools/docker/Dockerfile
@@ -40,7 +40,7 @@ COPY requirements/requirements_git.txt /tmp/requirements_git.txt
 RUN /opt/aap_gateway/venv/bin/pip --no-cache-dir install -r /tmp/requirements.txt
 RUN /opt/aap_gateway/venv/bin/pip --no-cache-dir install -r /tmp/requirements_git.txt
 
-FROM quay.io/centos/centos:stream9
+FROM quay.io/centos/centos:stream9 AS jewel
 ARG PYTHON
 
 RUN dnf module install -y nginx:1.22/common && \
@@ -91,11 +91,21 @@ COPY aap_gateway_api /opt/aap_gateway/src/aap_gateway_api/
 ENV PATH=/opt/aap_gateway/venv/bin:$PATH
 ENV PYTHONPATH=/opt/aap_gateway/src
 
-# Pull in the platform-ui, this can change frequently so it should be one of the last things we try and do
-COPY --from=quay.io/ansible/platform-ui:latest /usr/share/nginx/html /opt/aap_gateway/platform_ui
+# The platform UI is optional. The "jewel" target ships without it (used by CI
+# for the GHCR image). The "jewel-ui" target extends this stage with the UI.
+RUN mkdir -p /opt/aap_gateway/platform_ui && \
+    printf '%s\n' \
+      '<!DOCTYPE html><html><head><title>Jewel API</title></head><body>' \
+      '<h1>Jewel API Server</h1>' \
+      '<p>This image was built without the platform UI.</p>' \
+      '<p>Build with <code>--target jewel-ui</code> to include it.</p>' \
+      '</body></html>' > /opt/aap_gateway/platform_ui/index.html
 
 USER 1000
 EXPOSE 8000
 
 WORKDIR /opt
 CMD ["/usr/bin/launch-gateway"]
+
+FROM jewel AS jewel-ui
+COPY --from=quay.io/ansible/platform-ui:latest /usr/share/nginx/html /opt/aap_gateway/platform_ui


.github/workflows/build-image.yml
```diff
@@ -0,0 +1,62 @@
...
+  REGISTRY: ghcr.io
+  IMAGE_NAME: ansible/jewel
+
+jobs:
+  build-and-push-image:
+    if: github.repository == 'ansible/jewel'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
jay-steurer, line 22: **Critical — Supply chain risk: Actions pinned by mutable version tags**

All actions use mutable tags (`@v4`, `@v3`, etc.). If an upstream action repo is compromised, arbitrary code can be injected into this pipeline (ref: `tj-actions/changed-files` incident, March 2025).

Pin every action to its full commit SHA with a version comment:
```yaml
- uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
- uses: docker/login-action@74a5d142397b4f367a81961aca4a556e4ec26871 # v3.4.0
- uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
- uses: docker/setup-buildx-action@b5ca514318bd6ebac0fb2aedd5d36ec1b5c232a2 # v3.10.0
- uses: docker/metadata-action@902fa8ec7d6ecbf8d84d538b9b233a880e428804 # v5.7.0
- uses: docker/build-push-action@263435318d21b8e681c14492fe198571cfb76f00 # v6.18.0
```
(Verify exact SHAs against current releases.)

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Extract metadata (tags, labels) for Docker
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
+          tags: |
+            type=raw,value=latest
+            type=sha
+            type=sha,format=long
```
jay-steurer, line 53: **Important — `INCLUDE_UI=true` path is never tested in CI**

The workflow hardcodes `INCLUDE_UI=false` with no override mechanism. The `INCLUDE_UI=true` path is documented in the README but never exercised in CI — if the conditional copy logic breaks, no CI run will catch it.

Consider adding a `workflow_dispatch` trigger with an input parameter, or a second job that builds (no push) with `INCLUDE_UI=true` to validate both paths.

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v6
```
jay-steurer: **Suggestion — No smoke test before pushing to `latest`**

If a broken commit lands on `devel`, the `latest` tag gets overwritten with a broken image. There is no validation between build and push.

Consider a build-then-test-then-push pattern: build with `push: false` + `load: true`, run a quick smoke test (`docker run --rm jewel:test python -c "import aap_gateway_api"`), then rebuild multi-arch with `push: true`.

README.md
```diff
@@ -17,6 +17,29 @@ Jewel provides the source code for the Gateway that connects Ansible services an
 > The project name, Jewel, refers to a precious stone known as the Eye of the Sea from the same novel that originated the word Ansible, Ursula K. Le Guin's *Rocannon's World*.
 > Ansible communicates across distance; Jewel is what everything converges on.
 
+## Container Image
+
+A pre-built container image is published to the GitHub Container Registry on every push to the `devel` branch. The published image does **not** include the platform UI.
+
```
jay-steurer, line 23: **Important — README does not mention the published image excludes the platform UI**

A developer who pulls this image has no indication it ships without the UI. They will get a 404 in the browser with no explanation.

Suggested addition after the pull command:
> **Note:** The published image does not include the platform UI. To build an image with the UI included, see the local build instructions below.

tools/docker/Dockerfile
```diff
@@ -91,11 +91,13 @@ COPY aap_gateway_api /opt/aap_gateway/src/aap_gateway_api/
 ENV PATH=/opt/aap_gateway/venv/bin:$PATH
 ENV PYTHONPATH=/opt/aap_gateway/src
```
jay-steurer, line 93: **Suggestion — Add a comment explaining the conditional UI inclusion**

The old comment explaining why this block is near the end of the Dockerfile was removed. The new code is more complex and the staging approach is non-obvious. Suggested:
```dockerfile
# Platform UI inclusion is optional. CI images (GHCR) are built without it.
# Because COPY --from cannot be conditional, we stage files in /tmp and
# selectively copy them. Keep this near the end — the upstream image changes frequently.
```

tools/docker/Dockerfile
```diff
@@ -91,11 +91,13 @@ COPY aap_gateway_api /opt/aap_gateway/src/aap_gateway_api/
...
 
-# Pull in the platform-ui, this can change frequently so it should be one of the last things we try and do
-COPY --from=quay.io/ansible/platform-ui:latest /usr/share/nginx/html /opt/aap_gateway/platform_ui
+RUN mkdir -p /opt/aap_gateway/platform_ui
```
jay-steurer: **Critical — Empty `platform_ui` directory causes silent 404 for all browser requests**

When `INCLUDE_UI=false`, this directory exists but is empty. The nginx config (`tools/configs/nginx-gateway.conf:48-62`) serves from this path with `try_files $uri /index.html =404`. Since `index.html` doesn't exist, every browser request returns a raw 404 with no explanation.

Users pulling the GHCR image and opening it in a browser will think the container is broken.

Fix: add a placeholder `index.html` in the `else` branch that says the image is API-only and how to rebuild with `--build-arg INCLUDE_UI=true`.

tools/docker/Dockerfile
```diff
@@ -91,11 +91,13 @@ COPY aap_gateway_api /opt/aap_gateway/src/aap_gateway_api/
...
 
 USER 1000
```
jay-steurer, line 104: **Important — Platform-UI image always pulled even when `INCLUDE_UI=false`**

This `COPY --from` is unconditional — Docker pulls the entire `quay.io/ansible/platform-ui:latest` image on every build regardless of `INCLUDE_UI`. This means:
- Builds with `INCLUDE_UI=false` fail if quay.io is unreachable (defeating the purpose of making UI optional)
- The files leak into a recoverable image layer even after the `rm -rf`
- Unnecessary network traffic and build time

Since the workflow uses Buildx (BuildKit), consider `RUN --mount=from=` to avoid a persistent layer, or restructure as a multi-stage build with a conditional target.

Also: `platform-ui:latest` is a mutable tag — consider pinning to a digest for reproducible builds.

tools/docker/Dockerfile
```diff
@@ -91,11 +91,13 @@ COPY aap_gateway_api /opt/aap_gateway/src/aap_gateway_api/
...
 EXPOSE 8000
```
jay-steurer, line 105: **Important — `INCLUDE_UI` build arg has no validation; typos are silently ignored**

The check `[ "${INCLUDE_UI}" = "true" ]` is an exact string match. Values like `True`, `TRUE`, `yes`, `1`, or a typo like `ture` all silently evaluate to false — the UI is excluded with no warning.

Add validation:
```dockerfile
RUN if [ "${INCLUDE_UI}" != "true" ] && [ "${INCLUDE_UI}" != "false" ]; then \
      echo "ERROR: INCLUDE_UI must be 'true' or 'false', got '${INCLUDE_UI}'" && exit 1; \
    fi
```

jay-steurer (commented):
## PR Review Summary

**Critical (2)** · **Important (3)** · **Suggestions (4)**

See inline comments for details. The main blockers before merge are:
1. Empty `platform_ui` directory silently serves 404 to all browser requests
2. GitHub Actions pinned by mutable version tags (supply chain risk)

**Strengths:** Fork guard, scoped permissions, multi-platform build, accurate README tag table, clear workflow step names.

tools/docker/Dockerfile
```diff
@@ -91,11 +91,13 @@ COPY aap_gateway_api /opt/aap_gateway/src/aap_gateway_api/
...
```
jay-steurer, line 106: **Suggestion — `cp -r *` silently skips dotfiles (behavioral regression)**

The shell glob `*` does not match files beginning with `.` (e.g., `.well-known`, `.htaccess`). The original `COPY --from` copied everything including hidden files.

Fix: use `cp -a` with trailing dot syntax:
```dockerfile
cp -a /tmp/platform_ui/. /opt/aap_gateway/platform_ui/
```

.github/workflows/build-image.yml
Dostonbek1: Fixed — all actions are now pinned to full commit SHAs with version comments (e.g. `actions/checkout@11d5960a...  # v4.4.0`).

.github/workflows/build-image.yml
Dostonbek1: This was refactored — the `INCLUDE_UI` build arg no longer exists. The Dockerfile now uses named build targets: `jewel` (no UI, what CI publishes) and `jewel-ui` (extends `jewel` with a single `COPY --from`). The `jewel-ui` target is just one COPY instruction, so there's not much that can break independently.

.github/workflows/build-image.yml
Dostonbek1: Good idea — deferring for a follow-up. The main complication is that Buildx multi-arch builds can't `load` into local Docker, so we'd need to build twice (once single-arch for testing, once multi-arch for push). Worth adding once we have more experience with the pipeline.

README.md
Dostonbek1: This was already addressed in the previous push — the README now says: "The published image does **not** include the platform UI."

tools/docker/Dockerfile
Dostonbek1: Fixed — added a comment explaining the `jewel` / `jewel-ui` build target pattern and why UI inclusion is split into a separate stage.

tools/docker/Dockerfile
Dostonbek1: Great catch — fixed. The `jewel` target now writes a placeholder `index.html` that tells users this is an API-only image and how to rebuild with `--target jewel-ui` to include the UI.

tools/docker/Dockerfile
Dostonbek1: Already resolved — the Dockerfile was refactored to use named build targets instead of the `INCLUDE_UI` arg. When building with `--target jewel`, Docker never encounters the `COPY --from=quay.io/ansible/platform-ui` instruction (it's in the separate `jewel-ui` stage), so the image is never pulled and no layer is created.

tools/docker/Dockerfile
Dostonbek1: No longer applicable — the `INCLUDE_UI` build arg was removed entirely. The Dockerfile now uses named build targets (`jewel` vs `jewel-ui`), which eliminates the string validation issue.

tools/docker/Dockerfile
Dostonbek1: No longer applicable — the `cp -r` was removed. The `jewel-ui` stage uses `COPY --from` directly, which copies all files including dotfiles.

jay-steurer (approved):

.github/workflows/build-image.yml
```diff
@@ -0,0 +1,62 @@
...
+          persist-credentials: false
+
+      - name: Log in to the Container registry
+        uses: docker/login-action@c94ce9fb468520275223c153574b00df6fe4bcc9 # v3.7.0
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Set up QEMU
```
tznamena, line 38: Minor nitpick: it would be nice to have some comments here explaining wh y we need the QEMU and Buildx steps.

tznamena (approved):
I left a comment recommending to add comments but not those aren't blockers so approving.

.github/workflows/build-image.yml
Dostonbek1: Good call — added inline comments explaining that QEMU enables cross-platform emulation (arm64 on amd64 runners) and Buildx is needed for multi-arch builds with direct registry push.

Dostonbek1:
## Test Results

The workflow was tested end-to-end on a fork by temporarily adjusting the repo guard and image name:

- **Workflow run**: ✅ Completed successfully
- **Run URL**: https://github.com/Dostonbek1/jewel/actions/runs/30460913753
- **Image published**: `ghcr.io/dostonbek1/jewel:latest` (+ SHA tags)
- **Multi-arch**: Both `linux/amd64` and `linux/arm64` built and pushed
- **Cleanup**: Fork's `devel` branch restored, test branch deleted

tools/docker/Dockerfile
```diff
@@ -91,11 +91,21 @@ COPY aap_gateway_api /opt/aap_gateway/src/aap_gateway_api/
...
+      '</body></html>' > /opt/aap_gateway/platform_ui/index.html
 
 USER 1000
 EXPOSE 8000
 
 WORKDIR /opt
 CMD ["/usr/bin/launch-gateway"]
+
+FROM jewel AS jewel-ui
+COPY --from=quay.io/ansible/platform-ui:latest /usr/share/nginx/html /opt/aap_gateway/platform_ui
```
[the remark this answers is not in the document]
Dostonbek1: Intentionally keeping `:latest` here — the Jinja2 template (`Dockerfile.j2`) used by the local dev environment also references `platform-ui:latest`, and the UI image is updated frequently. Pinning to a digest would add ongoing maintenance burden to manually update it on every UI change. This only affects the optional `jewel-ui` target (not the published GHCR image, which uses the `jewel` target and never pulls this image).

PR #181, opened 2026-07-29
[AAP-82668] Fix resource sync blocking on org delete

john-westcott-iv:
## Summary

`AllServicesClient._make_request()` uses `ThreadPoolExecutor` as a context manager, which calls `shutdown(wait=True)` on exit — blocking until all service sync threads complete even when `wait_for_response=False`. This makes org deletes wait for controller/hub/EDA to process every cascaded resource delete.

## Fix

Use a module-level shared `ThreadPoolExecutor` for fire-and-forget requests. `submit()` returns immediately, the threads run in the background, and any delivery failures are caught by the 15-minute reverse sync.

## Performance Impact

Tested on scalelab with ~9000 users, ~3000 teams:

| Teams | Before | After |
|-------|--------|-------|
| 5     | 2.4s   | 0.4s  |
| 10    | 4.1s   | TBD   |
| 20    | 8.0s   | TBD   |
| 40    | timeout | TBD  |

## Risk

Low — the reverse sync (runs every 15 min in each service) catches any failed resource deletes. The `wait_for_response=True` path (used for create/update) is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resource synchronization “fire-and-forget” requests now return immediately, without waiting for downstream processing.
  * Background sync failures are isolated: timeouts and unexpected errors are caught so they don’t block the caller.
  * Synchronous waiting behavior remains unchanged, while background work no longer impacts normal operation or shutdown.

### DIFF
--- a/aap_gateway_api/utils/resources_client.py
+++ b/aap_gateway_api/utils/resources_client.py
@@ -2,6 +2,7 @@ import copy
 import logging
 import time
 from concurrent.futures import ThreadPoolExecutor, as_completed
+from threading import Lock
 from typing import Callable
 
 import requests
@@ -21,6 +22,25 @@ from aap_gateway_api.utils.preferences import get_preference_value
 ResourceRequestBody = DABResourceRequestBody
 
 logger = logging.getLogger('aap.gateway.utils.resource_api_client')
+
+# Lazily-initialized executor for fire-and-forget resource sync.
+# Created post-fork (on first use) to avoid sharing threads across uWSGI workers.
+_fire_and_forget_executor: ThreadPoolExecutor | None = None
+_executor_lock = Lock()
+
+
+def _get_executor() -> ThreadPoolExecutor:
+    global _fire_and_forget_executor
+
+    if _fire_and_forget_executor is None:
+        with _executor_lock:
+            if _fire_and_forget_executor is None:
+                _fire_and_forget_executor = ThreadPoolExecutor(
+                    max_workers=10,
+                    thread_name_prefix="resource-sync",
+                )
+
+    return _fire_and_forget_executor
 
 
 class GWResourceAPIClient(DABResourceAPIClient):
@@ -145,41 +165,46 @@ class AllServicesClient(GWResourceAPIClient):
                 raise
             return service.pk, None
 
-    # Executes requests in parallel using ThreadPoolExecutor to improve performance when
-    # making requests to multiple services simultaneously.
-    def _make_request(
-        self,
-        method: str,
-        path: str,
-        data: dict = None,
-        params: dict = None,
-    ) -> dict[int, Response | None]:
-        from aap_gateway_api.models import ServiceAPIRoute
+    def _make_async_request(self, services, method, path, data, params, jwt_token):
+        """Submit requests to the shared executor and return immediately.
 
+        Each request runs in a background thread. Errors are logged via
+        add_done_callback so they are never silently swallowed.
+        The 15-minute reverse sync in each service catches any delivery failures.
+        """
+        executor = _get_executor()
+        callback = self.callback
+        for svc in services:
+            future = executor.submit(self._async_worker, svc, method, path, data, params, jwt_token, callback)
+            future.add_done_callback(self._log_async_result)
+        return {}
+
+    @staticmethod
+    def _log_async_result(future):
+        """Log any unhandled exception from a fire-and-forget future."""
+        exc = future.exception()
+        if exc is not None:
+            logger.exception(f"Unhandled error in async resource sync: {exc}", exc_info=exc)
+
+    def _async_worker(self, service, method, path, data, params, jwt, callback):
+        """Execute a single service request in a background thread."""
+        try:
+            _, response = self._make_service_request(service, method, path, data, params, jwt)
+        except Timeout:
+            logger.error(f"Resource client request timeout for service {service.pk}")  # NOSONAR
+            response = None
+        except Exception as e:
+            logger.exception(f"Error in async request for service {service.pk}: {e}")
+            response = None
+        if callback:
+            try:
+                callback(service, response)
+            except Exception:
+                logger.exception(f"Callback failed for service {service.pk}")
+
+    def _make_synchronous_request(self, services, method, path, data, params, jwt_token):
+        """Execute requests in parallel and wait for all responses."""
         responses = {}
-        # Exclude gateway (not a downstream service) and services without a
-        # service-index endpoint (e.g. metrics) to avoid AttributeError in
-        # get_url_for_service when service_index_path is None or empty.
-        svc_qs = (
-            ServiceAPIRoute.objects.exclude(service_cluster__service_type__name=DefaultServiceType.GATEWAY.value)
-            .exclude(service_cluster__service_type__service_index_path__isnull=True)
-            .exclude(service_cluster__service_type__service_index_path='')
-        )
-        if self.service_filter:
-            svc_qs = svc_qs.filter(**self.service_filter)
-
-        # Evaluate queryset once to avoid lazy evaluation issues in threads
-        services = list(svc_qs)
-
-        # Early return if no services to process
-        if not services:
-            return responses
-
-        # Prefetch JWT to avoid race conditions with multiple threads calling refresh_jwt()
-        jwt_token = self.jwt
-
-        # Execute all service requests in parallel
-        # Cap max_workers to avoid excessive thread creation
         max_workers = min(len(services), 10)
         with ThreadPoolExecutor(max_workers=max_workers) as executor:
             futures = {executor.submit(self._make_service_request, svc, method, path, data, params, jwt_token): svc for svc in services}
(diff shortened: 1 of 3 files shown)


aap_gateway_api/utils/resources_client.py
```diff
@@ -145,41 +151,33 @@ def _make_service_request(self, service, method: str, path: str, data: dict = No
...
+        try:
+            _, response = self._make_service_request(service, method, path, data, params, jwt)
+        except Timeout:
+            logger.error(f"Resource client request timeout for service {service.pk}")  # NOSONAR
+            response = None
+        except Exception as e:
+            logger.exception(f"Error in async request for service {service.pk}: {e}")
+            response = None
+        if callback:
+            callback(service, response)
```
Dostonbek1: **Callback exception not observed.** If `callback` raises, the exception is silently swallowed since the future returned by `submit()` is never inspected. The synchronous path (line 196-197) has the same callback pattern but exceptions propagate naturally there.

Wrap the callback in a try/except:

```suggestion
        if callback:
            try:
                callback(service, response)
            except Exception:
                logger.exception(f"Callback failed for service {service.pk}")
```

Dostonbek1 (commented):
## Code Review Summary

Good fix for a real performance problem — the `ThreadPoolExecutor` context manager's `shutdown(wait=True)` was silently serializing the fire-and-forget path. The refactor cleanly separates async vs. sync execution and extracts service discovery into `_get_services()`. Two items worth addressing before merge.

**Files Reviewed**: 1 file
**Comments Posted**: 2 inline comments

### Issues Found
- 0 security/safety issues
- 1 correctness/stability issue (callback exception handling)
- 1 stability suggestion (DB connection lifecycle in long-lived executor threads)

### Overall Assessment
Needs minor fixes — the core approach is correct, but the two inline issues should be addressed for production stability.

### General Feedback
- The split into `_make_async_request` / `_make_synchronous_request` / `_get_services` is clean and makes the control flow much easier to follow.
- CodeRabbit's concern about the unbounded executor queue is theoretically valid but practically a non-issue: AAP deployments have 3-5 downstream services, so even burst deletes produce O(N) tasks where N is small. Not worth the complexity of bounded admission.
- The `_async_worker` Timeout catch (line 169) is technically unreachable since `_make_service_request` already catches Timeout and returns `None` when `wait_for_response=False` — but it's harmless defensive code, not worth removing.

aap_gateway_api/utils/resources_client.py
```diff
@@ -145,41 +151,33 @@ def _make_service_request(self, service, method: str, path: str, data: dict = No
...
+        The 15-minute reverse sync in each service catches any delivery failures.
+        """
+        callback = self.callback
+        for svc in services:
+            _FIRE_AND_FORGET_EXECUTOR.submit(self._async_worker, svc, method, path, data, params, jwt_token, callback)
+        return {}
 
-        # Execute all service requests in parallel
-        # Cap max_workers to avoid excessive thread creation
+    def _async_worker(self, service, method, path, data, params, jwt, callback):
```
Dostonbek1, line 189: **Stale DB connections in long-lived executor threads.** The module-level `_FIRE_AND_FORGET_EXECUTOR` keeps its 10 worker threads alive for the process lifetime. `get_url_for_service()` (called by `_make_service_request`) accesses `service.http_port` and `service.service_cluster.service_type`, which may trigger lazy DB queries. Without cleanup, these threads hold connections that can go stale between bursts.

The standard Django pattern for background threads:

```python
def _async_worker(self, service, method, path, data, params, jwt, callback):
    """Execute a single service request in a background thread."""
    from django.db import close_old_connections
    close_old_connections()
    try:
        ... # existing logic
    finally:
        close_old_connections()
```

Alternatively, adding `select_related('http_port', 'service_cluster__service_type')` to the queryset in `_get_services()` would eagerly load the related objects in the main thread, avoiding DB access in the worker threads entirely. That would also benefit the synchronous path.

aap_gateway_api/utils/resources_client.py
john-westcott-iv: Good catch. Wrapped the callback in try/except with logging. Added `test_callback_exception_is_logged` test.

aap_gateway_api/utils/resources_client.py
john-westcott-iv: Great point. Added `select_related('http_port', 'service_cluster__service_type')` to `_get_services()` so related objects are eagerly loaded in the main thread. This avoids DB access in worker threads entirely and benefits the synchronous path too.

aap_gateway_api/utils/resources_client.py
```diff
@@ -22,6 +22,12 @@
 
 logger = logging.getLogger('aap.gateway.utils.resource_api_client')
 
+# Shared executor for fire-and-forget resource sync requests.  submit()
+# returns immediately so the calling view (e.g. org/team delete) is not
+# blocked while downstream services process cascaded deletes.  Any delivery
+# failures are self-healing: each service runs a reverse sync every 15 min.
+_FIRE_AND_FORGET_EXECUTOR = ThreadPoolExecutor(max_workers=10)
+
```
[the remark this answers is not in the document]
john-westcott-iv: Acknowledged. The unbounded queue is a theoretical concern but not a practical risk in the current usage — org deletes are infrequent, each enqueues only one task per downstream service (3-4 services), and the 10 workers drain tasks in ~100ms each. A burst of 100 concurrent org deletes would queue ~400 tasks, which clears in seconds.

Adding a bounded queue would require a policy for rejected tasks (drop silently? log? block?), which adds complexity without a demonstrated need. The 15-minute reverse sync is the safety net for any delivery gaps.

If this becomes a real issue under load, we can revisit with a bounded queue + drop-and-log policy.

aap_gateway_api/utils/resources_client.py
```diff
@@ -22,6 +22,12 @@
 
 logger = logging.getLogger('aap.gateway.utils.resource_api_client')
 
+# Shared executor for fire-and-forget resource sync requests.  submit()
+# returns immediately so the calling view (e.g. org/team delete) is not
+# blocked while downstream services process cascaded deletes.  Any delivery
+# failures are self-healing: each service runs a reverse sync every 15 min.
+_FIRE_AND_FORGET_EXECUTOR = ThreadPoolExecutor(max_workers=10)
```
AlanCoding: This is only used in 1 location. I don't see why we need to have it initialized on import. That can cause a number of resource usage problems... maybe it doesn't. But still I don't see a reason to pay that cost.

AlanCoding: I believe my request was correct but my reason was wrong.

The issue isn't resource usage in general. The issue is having this created pre-fork where that will create resources shared across processes.

As a high level, "do we know what we're trying to do" kind of thing, we should check uWSGI for this setting. This is _desired_ for what we're trying to do. We don't want to share threads over multiple workers. We want to ad-hoc create threads in a worker. We don't want to bleed state between workers.

```
[uwsgi]
enable-threads = true
```

Then, within a worker, we still _want a global_ `ThreadPoolExecutor`. Similar to this, but _this_ does it pre-fork, which is death. So we want a _global_ but we want to lazily create it _at runtime_. So that would look like:

```python
_executor: ThreadPoolExecutor | None = None
_executor_lock = Lock()


def _get_executor() -> ThreadPoolExecutor:
    global _executor

    if _executor is None:
        with _executor_lock:
            if _executor is None:
                _executor = ThreadPoolExecutor(
                    max_workers=10,
                    thread_name_prefix="side-effect",
                )

    return _executor
```

aap_gateway_api/utils/resources_client.py
```diff
@@ -145,41 +151,36 @@ def _make_service_request(self, service, method: str, path: str, data: dict = No
...
-            return responses
-
-        # Prefetch JWT to avoid race conditions with multiple threads calling refresh_jwt()
-        jwt_token = self.jwt
+        Each request runs in a background thread. Errors are logged, not raised.
+        The 15-minute reverse sync in each service catches any delivery failures.
+        """
+        callback = self.callback
+        for svc in services:
+            _FIRE_AND_FORGET_EXECUTOR.submit(self._async_worker, svc, method, path, data, params, jwt_token, callback)
```
AlanCoding: you should still `add_done_callback` for the returned future here, and in that, just log the result. This shouldn't block concurrency

Dostonbek1 (commented):
## Re-Review — Updated Assessment

My [previous review comments](https://github.com/ansible/jewel/pull/181#pullrequestreview) are both addressed:
- **Callback exception handling** ✅ — `_async_worker` now wraps the callback in try/except with logging (line 175-178). Test coverage added (`test_callback_exception_is_logged`).
- **DB connections in long-lived threads** ✅ — `_get_services()` uses `select_related('http_port', 'service_cluster__service_type')` to eagerly load all relations in the main thread. Background threads don't touch the DB.

**Files Reviewed**: 3 files
**Comments Posted**: 0 new inline comments (existing review threads cover the remaining items)

### Remaining Items (from @AlanCoding)

1. **Pre-fork executor initialization (blocking)** — The module-level `_FIRE_AND_FORGET_EXECUTOR` is created at import time, which under uWSGI's prefork model means the executor (and its threads) are created in the master process before forking workers. Post-fork, child processes inherit the executor object but its threads belong to the dead parent — this causes silent failures or cross-process state corruption. Alan's suggested fix (lazy init with double-checked locking via `_get_executor()`) is the correct pattern and should be applied before merge.

2. **`add_done_callback` for observability (nice-to-have)** — While `_async_worker` already has comprehensive try/except, attaching `add_done_callback` to the future would catch edge cases where `submit()` itself fails (e.g., executor shutdown race) or where an exception escapes the try/except boundary. Low priority since the current error handling is solid, but easy to add.

### Overall Assessment

**Needs minor fix** — the refactoring is clean, tests are comprehensive (100% coverage on new code per Codecov/SonarCloud), and the core approach is correct. The pre-fork issue is the only blocking item. Once that's addressed with lazy initialization, this is good to merge.

_Assisted-by: Claude Code / Opus 4.6 (Anthropic)_

Dostonbek1 (approved):

aap_gateway_api/utils/resources_client.py
john-westcott-iv: Done. Replaced the module-level executor with lazy initialization via `_get_executor()` using double-checked locking. Created post-fork on first use to avoid sharing threads across uWSGI workers. Added `thread_name_prefix="resource-sync"` for debuggability.

Note: `enable-threads = true` is set in the dev uwsgi.ini but not in the RPM, containerized, or OCP installers. uWSGI >= 2.0.27 enables threads by default — we're checking which version ships.

aap_gateway_api/utils/resources_client.py
john-westcott-iv: Done. Added `future.add_done_callback(self._log_async_result)` after each submit. The callback checks `future.exception()` and logs any unhandled errors. Tests added for both the success and error cases.

aap_gateway_api/utils/resources_client.py
john-westcott-iv: Follow-up on the `enable-threads` concern: uWSGI 2.0.27 ([changelog](https://uwsgi-docs.readthedocs.io/en/latest/Changelog-2.0.27.html)) changed the default to `enable-threads = true`. The change is listed as:

> *core/uwsgi: set enable threads by default (Riccardo Magliocchetti)*

So any deployment running uWSGI >= 2.0.27 will have threading enabled without explicit config.

AlanCoding (approved):
All very much needed enhancements since a long time ago. All comments look to be addressed.

PR #189, opened 2026-08-03
[AAP-82668] Move deferral CMs inside perform_destroy for atomicity

AlanCoding:
Requires ansible/django-ansible-base#1076

## Summary
- Moves `defer_rbac_computations`, `deferred_activity_stream`, and `defer_resource_cleanup` inside `ResourceAPIUpdateMixin.perform_destroy` so the flush runs within `@transaction.atomic`
- Fixes the atomicity gap where org deletes could succeed while RBAC data remained inconsistent (flush failures now roll back the delete)
- All resource types benefit from deferral on delete, not just organizations
- Fixes test mock target (`compute_object_role_permissions` doesn't exist in triggers module)

## Background
Previously, deferral CMs were applied at the dispatch level in DAB's `AnsibleBaseView`, running OUTSIDE `@transaction.atomic`. The companion DAB PR reverts that dispatch wrapping. This PR applies the CMs where they belong — inside `perform_destroy`, wrapped by the existing `@transaction.atomic`.

The resource client call (`delete_resource`) stays outside the CMs since it's a remote call that shouldn't be replayed on retry.

Based on john_jewel/AAP-82668 which adds the atomicity and perf regression tests.

## Test plan
- [x] `TestOrgDeleteRBACFlushRollback::test_flush_failure_rolls_back_delete` passes
- [x] `TestOrgDeleteCleansUpRBAC::test_successful_delete_cleans_up_everything` passes
- [x] `test_org_delete_time_scales_sublinearly` perf test passes
- [x] Full org test file passes (24/24)

### DIFF
--- a/aap_gateway_api/tests/views/api/v1/organization/test_organization.py
+++ b/aap_gateway_api/tests/views/api/v1/organization/test_organization.py
@@ -1,7 +1,10 @@
+from unittest.mock import patch
+
 import pytest
 from ansible_base.lib.utils.response import get_relative_url
+from ansible_base.rbac.models import ObjectRole, RoleDefinition, RoleUserAssignment
 
-from aap_gateway_api.models import Organization
+from aap_gateway_api.models import Organization, Team
 
 
 def test_prevent_deletion_of_managed_organization(admin_api_client):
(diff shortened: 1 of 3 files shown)


jay-steurer:
@AlanCoding we have Tier1 test cases failing, we will need manual validation on those tests since konflux currently don't update the DAB version to run Tier1.

AlanCoding:
That makes sense. Even the CI here was failing before adding the `Requires https://github.com/ansible/django-ansible-base/pull/1076` syntax. It very intentionally raises an exception.

john-westcott-iv:
Closing in favor of #1047

PR #194, opened 2026-08-06
AAP-85286: Fix _is_service_already_synced() to detect empty registry

jay-steurer:
## Summary

- Fixes `_is_service_already_synced()` to distinguish "empty registry" from "already synced"
- Adds an unfiltered `list_resources(filters={})` query before the per-type checks — if total count is 0, the registry is empty (not populated yet), so migration should proceed rather than being skipped
- Prevents the false "already synchronized" short-circuit that left users/teams/orgs unmigrated after 2.4-to-2.6 OCP upgrades
- 
Requires ansible/django-ansible-base#1091 

## Problem

After OCP upgrades, the controller's resource registry can be empty because DAB's `initialize_resources` backfill was skipped during multi-pass migrations. The `_is_service_already_synced()` method queries each resource type with `service_id` and `is_partially_migrated` filters — when all return `count: 0`, it concludes "already synced" and skips the entire migration. It cannot distinguish between:
- "All resources migrated" (count=0 because they moved to gateway's service_id)
- "Registry is empty" (count=0 because `initialize_resources` never ran)

## Changes

- **Modified:** `aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py` — added unfiltered query with WARNING log when registry is empty
- **Modified:** `aap_gateway_api/tests/management/commands/_migrate_service_data/test_service_orchestration.py` — updated 2 existing tests for the new unfiltered query, added 1 new test

## Companion PR

- DAB: ansible/django-ansible-base#1091 — adds `manage.py initialize_resource_registry` command to fix the root cause (backfill never running)

## Test plan

- [x] `test_migration_does_not_skip_when_registry_empty` — empty registry (total=0) returns False with warning
- [x] `test_migration_skips_when_already_synced` — populated registry, all migrated (total>0, filtered=0) returns True
- [x] `test_migration_proceeds_when_not_synced` — unmigrated resources (filtered>0) returns False
- [x] `test_migration_uses_bulk_fetch` — updated to account for new unfiltered query
- [x] Full `_migrate_service_data` test suite passes (173/174, 1 pre-existing flaky benchmark)

Assisted-by: Claude Code

## Bug Fixes
- Improved service migration checks when no upstream resources are available.
- Migrations now proceed with a warning when the upstream registry is empty.
- Refined handling of system-user-only and unsupported resources.
- Improved detection of remaining resources across configured resource types.

## Tests
- Added coverage for empty and populated registries.
- Expanded validation for supported, unsupported, and system-only resources.

### DIFF
--- a/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
+++ b/aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
@@ -11,16 +11,30 @@ from django.db import transaction
 
 class ResourceMigrationMixin:
     def _is_service_already_synced(self) -> bool:
-        """Check if all resource types for the current service have already been migrated."""
-        for resource_type_name in self.resource_types_to_migrate:
-            filters = {
+        """Check if all migratable resource types for the current service have already been migrated."""
+        response = self.client.list_resources(
+            filters={
                 "service_id": self.upstream_service_id,
                 "is_partially_migrated": "false",
-                "content_type__resource_type__name": resource_type_name,
+                "content_type__resource_type__name__in": ",".join(self.resource_types_to_migrate),
             }
-            data = self.client.list_resources(filters=filters).json()
-            if data["count"] > 0:
+        ).json()
+
+        if response["count"] > 0:
+            results = response.get("results", [])
+            if settings.SYSTEM_USERNAME:
+                results = [r for r in results if r.get("resource_type") != SHARED_USER_RESOURCE_TYPE or r.get("name") != settings.SYSTEM_USERNAME]
+            if results or response["count"] > len(response.get("results", [])):
                 return False
+
+        has_resources = self.client.list_resources(filters={"service_id": self.upstream_service_id}).json()
+        if has_resources["count"] == 0:
+            self._log(
+                "Upstream resource registry is empty — will attempt migration (registry may not be populated yet).",
+                logging.WARNING,
+            )
+            return False
+
         return True
 
     def update_resource_data(self, resource_type_name: str, original_resource_data: Any) -> Optional[Dict[str, Any]]:
--- a/aap_gateway_api/tests/management/commands/_migrate_service_data/test_service_orchestration.py
+++ b/aap_gateway_api/tests/management/commands/_migrate_service_data/test_service_orchestration.py
@@ -5,7 +5,9 @@ from collections import OrderedDict
 from unittest.mock import Mock, patch
 
 import pytest
+from django.conf import settings
 from django.core.management import call_command
+from django.core.management.base import CommandError
 
 from aap_gateway_api.management.commands.migrate_service_data import Command as MigrateCommand
 from aap_gateway_api.tests.management.commands._migrate_service_data.conftest import SEP_CHAR, assert_all_resources_synced, setup_empty_assignment_mocks
@@ -179,10 +181,21 @@ def test_migration_skips_when_already_synced(admin_user, capsys, service_api_rou
                 "service_id": str(uuid.uuid4()),
                 "service_type": "controller",
             }
-            mock_client.list_resources.return_value.json.return_value = {
-                "count": 0,
-                "results": [],
-            }
+
+            def list_resources_side_effect(filters=None):
+                resp = Mock()
+                filters = filters or {}
+                if "is_partially_migrated" in filters:
+                    # Unmigrated query: no unmigrated resources remain
+                    resp.json.return_value = {"count": 0, "results": []}
+                elif "service_id" in filters:
+                    # All-resources query: registry has resources (they're all migrated)
+                    resp.json.return_value = {"count": 5, "results": []}
+                else:
+                    resp.json.return_value = {"count": 0, "results": []}
+                return resp
+
+            mock_client.list_resources.side_effect = list_resources_side_effect
             setup_empty_assignment_mocks(mock_client)
             return mock_client
 
@@ -195,6 +208,44 @@ def test_migration_skips_when_already_synced(admin_user, capsys, service_api_rou
         assert "skipping resource migration" in captured.out
         assert "Migrating data for" not in captured.out
         assert "role assignments" in captured.out
+
+
+@pytest.mark.django_db(transaction=True)
+def test_migration_does_not_skip_when_registry_empty(admin_user, capsys, service_api_route_controller, patched_resource_client, system_user):
+    """When upstream registry is completely empty (not populated yet), migration should NOT be skipped."""
+
+    with (
+        patch('aap_gateway_api.utils.resources_client.GWResourceAPIClient') as mock_client_class,
+        patch('aap_gateway_api.management.commands.migrate_service_data.Command.load_types_and_permissions'),
+    ):
+
+        def mock_client_factory(service_api, *args, **kwargs):
+            mock_client = Mock()
+            mock_client.service = service_api
+            mock_client.user = admin_user
+            mock_client.get_service_metadata.return_value.json.return_value = {
+                "service_id": str(uuid.uuid4()),
+                "service_type": "controller",
+            }
+            # All queries return count=0 — registry is empty
+            mock_client.list_resources.return_value.json.return_value = {
+                "count": 0,
+                "results": [],
+            }
+            setup_empty_assignment_mocks(mock_client)
+            return mock_client
+
+        mock_client_class.side_effect = mock_client_factory
+
+        call_command("migrate_service_data", username=admin_user.username)
+
+        captured = capsys.readouterr()
+        # Should NOT say "already synchronized"
+        assert "already synchronized" not in captured.out
+        # Should warn about empty registry
+        assert "empty" in captured.err.lower() or "empty" in captured.out.lower()
+        # Should attempt migration
+        assert "Migrating data for" in captured.out
 
 
 @pytest.mark.django_db(transaction=True)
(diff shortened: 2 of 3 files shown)


john-westcott-iv (changes requested):
The fix correctly identifies the empty-registry problem, but I think the overall logic can be simplified.

The current approach makes **O(N+1) API calls** — 1 unfiltered query + 1 per resource type. We can collapse this to **O(≤2)** (that's a real complexity class, I promise) by dropping the per-type loop entirely:

```python
def _is_service_already_synced(self) -> bool:
    needs_migration_response = self.client.list_resources(
        filters={
            "service_id": self.upstream_service_id,
            "is_partially_migrated": "false",
        }
    ).json()
    if needs_migration_response["count"] > 0:
        return False

    has_resources = self.client.list_resources(
        filters={"service_id": self.upstream_service_id}
    ).json()
    if has_resources["count"] == 0:
        self._log(
            "Upstream resource registry is empty — will attempt migration "
            "(registry may not be populated yet).",
            logging.WARNING,
        )
        return False

    return True
```

**Why this is better:**

1. **Fewer API calls** — at most 2, regardless of how many resource types exist, vs 1 + N today
2. **Simpler logic flow** — the three states are cleanly separated: unmigrated resources exist → migrate; no resources at all → warn and migrate; resources exist but all migrated → skip
3. **The empty-registry check moves to where it matters** — right before the `return True`, which is the only place where "empty" and "all migrated" would be confused. The current PR checks emptiness first (before we even know if it's relevant)

All resource types use the same `upstream_service_id` and the same `is_partially_migrated` filter, so aggregating across types should be equivalent to looping per-type.

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -11,16 +11,38 @@
...
+                    "service_id": self.upstream_service_id,
+                    "is_partially_migrated": "false",
+                    "content_type__resource_type__name": resource_type_name,
+                }
+            ).json()
+            count = response["count"]
+
+            if count > 0 and resource_type_name == SHARED_USER_RESOURCE_TYPE and settings.SYSTEM_USERNAME:
+                results = response.get("results", [])
+                if len(results) == count and all(r.get("name") == settings.SYSTEM_USERNAME for r in results):
```
Dostonbek1: The system-user exclusion here uses a different strategy than `_get_filtered_resources()` (lines 221–226), which filters `results` in-place. This `len(results) == count` guard is fragile when the API paginates — if `count > 0` but the first page of `results` is empty or incomplete, the guard will return `False` and re-run migration even though only the system user remains unmigrated.

Consider aligning with the migration path: filter system users from `results` and check whether any entries remain, rather than requiring `len(results) == count`:

```suggestion
                results = [r for r in response.get("results", []) if r.get("name") != settings.SYSTEM_USERNAME]
                if not results:
                    continue
```

This matches `_get_filtered_resources` semantics without pulling in `RESOURCE_DATA_FILTERS` (keeping the guard query lightweight).

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
```diff
@@ -11,16 +11,38 @@
...
 class ResourceMigrationMixin:
     def _is_service_already_synced(self) -> bool:
-        """Check if all resource types for the current service have already been migrated."""
+        """Check if all migratable resource types for the current service have already been migrated.
+
+        Scopes queries to resource_types_to_migrate so that unsupported resource
+        types in the registry do not prevent the guard from short-circuiting.
+        For shared.user, the system user is excluded to match migrate_resource().
+        """
         for resource_type_name in self.resource_types_to_migrate:
```
Dostonbek1: This per-type loop makes **N+1 API calls** (one per entry in `resource_types_to_migrate` plus the empty-registry check). Since `content_type__resource_type__name__in` is already used elsewhere in this test suite (`conftest.py`), you could collapse the unmigrated check to a single scoped query:

```python
self.client.list_resources(filters={
    "service_id": self.upstream_service_id,
    "is_partially_migrated": "false",
    "content_type__resource_type__name__in": ",".join(self.resource_types_to_migrate),
})
```

Then apply the system-user filter on `shared.user` results from that one response. This addresses @john-westcott-iv's O(≤2) suggestion while preserving the scoping that CodeRabbit requested.

aap_gateway_api/tests/management/commands/_migrate_service_data/test_service_orchestration.py
```diff
@@ -230,6 +280,97 @@ def mock_client_factory(service_api, *args, **kwargs):
...
+            mock_client.user = admin_user
+            mock_client.get_service_metadata.return_value.json.return_value = {
+                "service_id": str(uuid.uuid4()),
+                "service_type": "controller",
+            }
+
+            def list_resources_side_effect(filters=None):
+                resp = Mock()
+                filters = filters or {}
+                if "content_type__resource_type__name" in filters:
```
Dostonbek1: This mock returns `count: 0` for all per-type queries, so it doesn't actually exercise the scenario described in the docstring — unmigrated resources of an **unsupported** type blocking the guard. A stronger test would simulate an aggregated query returning `count > 0` for unmigrated resources of a type *not* in `resource_types_to_migrate` (e.g. `controller.jobtemplate`), while per-type scoped queries return `count: 0`. That would fail with a naive single-query guard but pass with this per-type loop, proving the scoping matters.

Dostonbek1 (commented):
## Code Review Summary

This PR correctly fixes the empty-registry false positive in `_is_service_already_synced()` — the core bug that caused migration to be skipped after OCP upgrades when `initialize_resources` never ran. The empty-registry check (`service_id`-scoped count == 0 → warn and proceed) is well-placed after the per-type unmigrated checks, and the added tests cover the main scenarios including system-user-only and already-synced paths.

**Files Reviewed**: 2 files
**Comments Posted**: 3 review comments

### Issues Found
- 0 security/safety issues
- 1 correctness/logic issues (pagination edge case in system-user guard)
- 2 code quality suggestions (API call consolidation, test coverage gap)

### Overall Assessment
Needs minor fixes — the bug fix is sound and well-tested for the primary regression, but the system-user exclusion logic has a pagination edge case and could be simplified by reusing existing filtering patterns.

### General Feedback
- The latest commit appropriately addresses CodeRabbit's scoping feedback (per-type loop + system-user exclusion) while keeping John's O(2) optimization as a viable follow-up via `content_type__resource_type__name__in`.
- Test coverage is solid for the happy paths; `test_migration_guard_ignores_unsupported_resource_types` could be strengthened to actually assert the unsupported-type scenario.
- Consider a direct unit test on `_is_service_already_synced()` (vs. full `call_command` integration) for faster, more targeted guard coverage.

jay-steurer:
@john-westcott-iv Good call on the O(≤2) simplification. I collapsed the per-type loop into a single scoped query using `content_type__resource_type__name__in`, so it's now at most 2 API calls regardless of how many resource types exist.

I kept the type scoping (rather than a fully unscoped unmigrated check) because without it, unsupported resource types that are permanently unmigrated would prevent the guard from ever returning `True`. But the flow now matches the 3-state structure you suggested:

1. Unmigrated migratable resources exist → migrate
2. Registry empty → warn and migrate
3. All migratable resources migrated → skip

Also switched the system-user exclusion from `len(results) == count` to a simple list filter, as @Dostonbek1 pointed out the pagination fragility.

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
jay-steurer: Done — collapsed to a single `content_type__resource_type__name__in` query, now O(≤2) total. Thanks for the suggestion, this is cleaner than the per-type loop.

aap_gateway_api/management/commands/_migrate_service_data/resource_migration.py
jay-steurer: Good catch on the pagination edge case. Replaced with a simple list filter:
```python
results = [r for r in results if r.get("name") != settings.SYSTEM_USERNAME]
```
This matches `_get_filtered_resources` semantics without the `len == count` assumption.

aap_gateway_api/tests/management/commands/_migrate_service_data/test_service_orchestration.py
jay-steurer: Agreed. Updated the test to simulate concrete unsupported types (`controller.jobtemplate`) in the unscoped response with `count: 5`, while the scoped unmigrated query returns `count: 0`. This proves the scoping matters — a naive unscoped guard would see those unmigrated resources and never skip.

Dostonbek1 (approved):
## Code Review Summary

Re-reviewed after the latest commit (`cf1383d`). All prior feedback has been addressed: the per-type loop is collapsed to a single scoped `content_type__resource_type__name__in` query (O(≤2) API calls), system-user exclusion is scoped to `shared.user` via `resource_type`, and the pagination heuristic (`count > len(results)`) conservatively returns `False` when additional pages may contain unmigrated resources.

**Files Reviewed**: 2 files
**Comments Posted**: 0 review comments

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 0 code quality suggestions

### Overall Assessment
LGTM

### General Feedback
- The empty-registry fix is correct and well-tested — the core regression (false "already synchronized" skip after OCP upgrades) is resolved.
- Test coverage is thorough: empty registry, already-synced, not-synced, system-user-only, org-name-collision-with-system-username, and unsupported resource types.
- The `count > len(results)` guard on line 27 handles the pagination edge case CodeRabbit flagged — when the first page is all system users but more pages exist, migration correctly proceeds.
- Resolves @john-westcott-iv's simplification request while preserving CodeRabbit's scoping requirements.

jay-steurer:
Since @john-westcott-iv  is not available, I will skip is approval. Please reach out if you have any concerns.

PR #197, opened 2026-08-10
AAP-87035 Fix pr-labels workflow failing on Node 24

jay-steurer:
## Summary
- Replace `require('@octokit/rest')` with native `fetch()` for the org membership check in the pr-labels workflow
- The `@octokit/rest` module is not bundled with `actions/github-script` and fails to load on Node 24

## Changes Made
- Switched from `Octokit` client to `fetch()` API for the GitHub org membership check
- Simplified error handling: 302 redirects (non-org-member requester) are now handled as a normal response status rather than a caught exception
- Removed the `@octokit/rest` dependency entirely from the workflow

## Test Plan
- [ ] Verify the pr-labels workflow runs successfully on a PR
- [ ] Confirm org membership detection works for ansible org members
- [ ] Confirm external contributors are correctly identified

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved community membership verification for more reliable results.
  * Successful membership responses are now recognized correctly, while redirected responses are handled safely.
  * Unexpected verification responses and connection errors now generate warnings, improving visibility into membership-check issues.

### DIFF
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -284,8 +284,9 @@ jobs:
                   isInternal = true;
                   core.info(`${pr.user.login} is a member of the ansible org`);
                 } else if (resp.status === 302) {
-                  isInternal = true;
-                  core.info(`${pr.user.login} is a member of the ansible org (requester is not org member)`);
+                  core.warning(`Org membership check returned 302 — ORG_TOKEN owner is not an ansible org member; cannot determine membership for ${pr.user.login}`);
+                } else if (resp.status !== 404) {
+                  core.warning(`Unexpected status ${resp.status} checking org membership for ${pr.user.login}`);
                 }
               } catch (e) {
                 core.warning(`Failed to check org membership for ${pr.user.login}: ${e.message}`);
--- a/.github/workflows/pr-labels.yml
+++ b/.github/workflows/pr-labels.yml
@@ -276,23 +276,19 @@ jobs:
 
             if (!isInternal) {
               try {
-                const { Octokit } = require('@octokit/rest');
-                const orgOctokit = new Octokit({ auth: process.env.ORG_TOKEN });
-                const { status } = await orgOctokit.orgs.checkMembershipForUser({
-                  org: 'ansible',
-                  username: pr.user.login,
-                });
-                if (status === 204) {
+                const resp = await fetch(
+                  `https://api.github.com/orgs/ansible/members/${pr.user.login}`,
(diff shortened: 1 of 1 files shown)


jay-steurer:
Addressed CodeRabbit's review comment in 28e8664: HTTP 302 now warns instead of setting `isInternal = true`, since 302 means the ORG_TOKEN owner is not an org member (not that the target user is). Also added a warning for other unexpected status codes (anything other than 204/302/404).

Dostonbek1 (approved):
## Code Review Summary

Clean fix for the Node 24 compatibility issue — replaces `require('@octokit/rest')` (not bundled in `actions/github-script`) with native `fetch()`. The second commit correctly addresses the HTTP 302 semantics per GitHub API docs: 302 means the *requester* isn't an org member, not that the target user *is* — so warning instead of assuming internal is the right call.

**Files Reviewed**: 1 file
**Comments Posted**: 0 inline comments

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 0 code quality suggestions

### Overall Assessment
LGTM

### General Feedback
- `redirect: 'manual'` correctly captures the 302 without following it
- The status handling hierarchy (204 → internal, 302 → unknown/warn, 404 → not a member, else → warn) properly reflects the GitHub org membership API semantics
- The `USERNAME_MAPPING` fallback still provides a safety net for edge cases where the org check can't determine membership
- CodeRabbit suggested a `membershipKnown` flag to skip Community-label mutations on unknown results — that's a reasonable follow-up but not a blocker for this fix

.github/workflows/pr-labels.yml
```diff
@@ -276,23 +276,20 @@ jobs:
...
               try {
-                const { Octokit } = require('@octokit/rest');
-                const orgOctokit = new Octokit({ auth: process.env.ORG_TOKEN });
-                const { status } = await orgOctokit.orgs.checkMembershipForUser({
-                  org: 'ansible',
-                  username: pr.user.login,
-                });
-                if (status === 204) {
+                const resp = await fetch(
+                  `https://api.github.com/orgs/ansible/members/${pr.user.login}`,
```
john-westcott-iv, line 280: I've had issues with this in the past as you might not see an org member if their membership status is private instead of public. Something to be aware of but this isn't a divergence from the initial behavior.

john-westcott-iv (approved):
Clean fix for the Node 24 compatibility issue. HTTP status handling is correct per the GitHub API docs, and the second commit properly addresses the 302 semantics. LGTM.

PR #202, opened 2026-08-17
[AAP-88288] Fix rotate_secret_key for JSONField (jsonb) encrypted columns

amasolov:
## Description

- **What**: Fixes `rotate_secret_key` to correctly handle encrypted fields stored in PostgreSQL `jsonb` columns (specifically `AuthenticatorUser.extra_data`).
- **Why**: The command silently skipped jsonb-backed encrypted fields during rotation, leaving them encrypted with the old key. After rotation, user authentication fails with `InvalidToken` errors.
- **How**: Detects JSONField columns, normalises the raw value returned by psycopg (unwrapping JSON string encoding if present), and uses `::jsonb` cast when writing back.

Closes AAP-88288

## Type of Change

- [x] Test update

## Self-Review Checklist

### Prerequisites
- PostgreSQL running (via podman or tox-docker)
- Python 3.12 venv with jewel dependencies installed

### Steps to Test
1. Install deps and run the full test suite:
   ```
   tox -e py312 -- aap_gateway_api/tests/management/test_rotate_secret_key.py -v
   ```
2. Specifically validate the new jsonb tests:
   ```
   pytest aap_gateway_api/tests/management/test_rotate_secret_key.py -v -k jsonb
   ```
3. For live testing: deploy AAP 2.7.4, authenticate users (creating AuthenticatorUser records), run `aap-gateway-manage rotate_secret_key`, verify users can still authenticate.

### Expected Results
- All existing tests continue to pass (no regression)
- 3 new tests pass: `test_jsonb_encrypted_field_rotation`, `test_jsonb_encrypted_field_with_json_wrapped_value`, `test_normalise_jsonb_value`
- AuthenticatorUser.extra_data is correctly re-encrypted during rotation (no longer skipped)
- User authentication works after rotation without needing to delete AuthenticatorUser records

### Root cause detail
`AuthenticatorUser.extra_data` is a `JSONField` (jsonb) with `encrypted_fields = ["extra_data"]`. When the ORM saves it, the dict value is encrypted into a string `$encrypted$UTF8$AESCBC$...`, which Django's JSONField stores as a JSON string in the jsonb column.

When reading via raw SQL `connection.cursor()`, psycopg may return the value still JSON-wrapped with outer quotes (`"$encrypted$..."`) depending on adapter configuration. The previous code assumed all values would be plain strings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Secret key rotation now correctly handles encrypted values stored in JSONB fields.
  * Nested JSON data remains intact and readable after re-encryption.
  * Improved support for raw and JSON-wrapped ciphertext, including normalization of malformed or unsupported values.
  * Empty, non-string, and unencrypted values are safely skipped during rotation.

* **Tests**
  * Added coverage for JSONB encryption, decryption, rotation, nested payload preservation, and unsupported data formats.


amasolov:
## Live Testing on AAP 2.7.4 (Operator deployment, OpenShift)

Tested the fix against a running AAP 2.7.4 Gateway deployment with 3 `AuthenticatorUser` records containing encrypted `extra_data` (jsonb columns).

**Shipped (unpatched) command:**
```
$ GATEWAY_SECRET_KEY="..." aap-gateway-manage rotate_secret_key --use-custom-key --dry-run

WARNING: Cannot decrypt AuthenticatorUser.extra_data pk=2 with the current SECRET_KEY; skipping
WARNING: Cannot decrypt AuthenticatorUser.extra_data pk=3 with the current SECRET_KEY; skipping
WARNING: Cannot decrypt AuthenticatorUser.extra_data pk=4 with the current SECRET_KEY; skipping
WARNING: 3 value(s) could not be decrypted and were NOT re-encrypted.
7 value(s) would be re-encrypted.
```

**Patched command (this PR):**
```
$ GATEWAY_SECRET_KEY="..." aap-gateway-manage rotate_secret_key --use-custom-key --dry-run

10 value(s) would be re-encrypted.
Preference cache would be flushed.
```

**Full rotation result:**
- 10 values re-encrypted (7 encrypted fields + 3 preferences), 0 skipped
- Gateway restarted successfully with the new key
- Post-rotation authentication: admin (200), test user (200)
- All `AuthenticatorUser.extra_data` fields decrypt correctly with the new key
- No data loss, no errors

amasolov:
@coderabbitai review

john-westcott-iv (changes requested):
Good fix — the root cause is well-identified, the live testing on AAP 2.7.4 is strong, and the test coverage is thorough. A few structural suggestions to simplify the code:

### 1. Inline the JSONField check, drop `_is_json_field`

`_is_json_field` is a 6-line static method with a deferred import for a single `isinstance` call. Import `JSONField` at the top of the file and inline it:

```python
from django.db.models import JSONField

# then in _rotate_encrypted_fields:
is_jsonb = isinstance(field_obj, JSONField)
```

### 2. Separate concerns: normalization, crypto, and serialization

Right now `is_jsonb` is threaded through 4 method calls (`_rotate_encrypted_fields` → `_reencrypt_column` → `_paginated_reencrypt` → `_reencrypt_value`), and `_reencrypt_value` handles both the jsonb normalization on input and `json.dumps` on output. These are three distinct concerns that should live in their own layers:

- **Input normalization** — unwrap what psycopg gave us (call unconditionally, no flag needed)
- **Crypto** — decrypt old key, encrypt new key, return ciphertext (format-agnostic, no `is_jsonb`)
- **Output serialization** — `json.dumps` + `::jsonb` cast for the SQL write (only place `is_jsonb` lives)

With this separation:
- `_reencrypt_value` loses its `is_jsonb` parameter entirely — it just gets a string in and returns a string out
- `_paginated_reencrypt` loses `is_jsonb` too — it calls `_normalise_if_jsonb_value` unconditionally before `_reencrypt_value`, and handles `json.dumps` wrapping at the SQL execution point where `is_jsonb` is already known
- `is_jsonb` only survives in `_reencrypt_column`, where it's passed to `_build_column_update_sql` (for the `::jsonb` cast) and used at the UPDATE execution site (for `json.dumps`)

The SQL execution in `_paginated_reencrypt` would become:

```python
for pk, raw in rows:
    normalised = self._normalise_if_jsonb_value(raw)
    new_val = self._reencrypt_value(normalised, label=f"{label} pk={pk}")
    if new_val is None:
        continue
    if not dry_run:
        # Output serialization lives at the write boundary
        write_val = json.dumps(new_val) if is_jsonb else new_val
        cursor.execute(update_sql, [write_val, pk])
    count += 1
```

And `_build_column_update_sql` owns the `::jsonb` cast:

```python
@staticmethod
def _build_column_update_sql(model, column_name, *, is_jsonb=False):
    qn = connection.ops.quote_name
    cast = "::jsonb" if is_jsonb else ""
    return "UPDATE {table} SET {col} = %s{cast} WHERE {pk} = %s".format(
        table=qn(model._meta.db_table),
        col=qn(column_name),
        pk=qn(model._meta.pk.column),
        cast=cast,
    )
```

The `json.dumps` and `::jsonb` cast are co-located at the write boundary — they're always needed together, and nowhere else.

### 3. Simplify `_normalise_jsonb_value` — rename and remove `ENCRYPTED_STRING` awareness

The current implementation special-cases `ENCRYPTED_STRING` at the top, but the calling function already checks for it. The normalizer's job should be pure JSON unwrapping.

Also, since the function is a safe no-op for non-jsonb values (plain string in → plain string out, json.loads fails → returns raw), it can be called unconditionally without an `if is_jsonb` guard. Rename to `_normalise_if_jsonb_value` to make that self-documenting:

```python
@staticmethod
def _normalise_if_jsonb_value(raw):
    """Unwrap a value that may be JSON-encoded into a plain Python string.

    Safe to call on non-jsonb values — a plain string that isn't valid
    JSON is returned as-is.
    """
    if not isinstance(raw, str):
        return None
    try:
        parsed = json.loads(raw)
        if isinstance(parsed, str):
            return parsed
    except (json.JSONDecodeError, TypeError):
        pass
    return raw
```

Flow:
- Not a string (dict, list, None) → `None`
- JSON-wrapped string (`'"$encrypted$..."'`) → json.loads → unwrapped `$encrypted$...`
- Already-unwrapped string (`$encrypted$...`) → json.loads fails → return raw
- Any other string → same treatment, caller checks for `ENCRYPTED_STRING` downstream

amasolov:
@john-westcott-iv Thank you for the review! The changes have been implemented as requested.

john-westcott-iv (approved):
Re-reviewed — all previous comments addressed cleanly.

- **JSONField check inlined** — `_is_json_field` removed, `isinstance(field_obj, JSONField)` at the top with top-level import ✔
- **Concerns separated** — `_normalise_if_jsonb_value()` called unconditionally (pure JSON unwrapping, no `ENCRYPTED_STRING` awareness), `_reencrypt_value()` is format-agnostic (no `is_jsonb`), `json.dumps` + `::jsonb` at the write boundary ✔
- **Normalizer simplified** — matches proposed implementation exactly, renamed to `_normalise_if_jsonb_value` ✔
- **Bonus `_reencrypt_rows` extraction** — clean cognitive complexity reduction ✔
- **Tests updated** — reflect the new two-step flow (normalise then reencrypt), no `is_jsonb` on `_reencrypt_value` ✔
- **CI** — all 23 checks passing ✔

**Non-blocking follow-up suggestion:** `_reencrypt_rows` currently opens a cursor and executes a single UPDATE per row — 100 cursor open/close/execute cycles per page. Consider switching to `executemany` to batch the updates into one cursor call per page:

```python
updates = []
for pk, raw in rows:
    normalised = self._normalise_if_jsonb_value(raw)
    new_val = self._reencrypt_value(normalised, label=f"{label} pk={pk}")
    if new_val is None:
        continue
    count += 1
    if not dry_run:
        write_val = json.dumps(new_val) if is_jsonb else new_val
        updates.append((write_val, pk))
if updates:
    with connection.cursor() as ucur:
        ucur.executemany(update_sql, updates)
```

This is a key rotation command that runs during maintenance windows, so it's not urgent — but it's the architecturally correct pattern and would matter for large datasets.

LGTM.

jay-steurer (approved):
Fix has been validated. Merging it

PR #220, opened 2026-08-26
[AAP-89439] Restore DABCacheWithFallback as default cache backend

tznamena:
Requires: https://github.com/ansible/ansible.platform/pull/241

## Description

- **What is being changed?**
  The default Django cache backend (`CACHES["default"]`) is switched back from `DABRedisCache` (Unix socket) to `DABCacheWithFallback`. The `DABRedisCache` configuration is preserved under the `"redis_local"` alias for sidecar deployments to opt into.

- **Why is this change needed?**
  Commit `af8d990c` (AAP-65921) hardcoded a Unix socket path (`/var/run/redis/redis.sock`) as the default cache location. On operator deployments where Redis runs as a separate pod accessible via TCP, this socket does not exist. Because `DABRedisCache` silently swallows `ConnectionError` via `DJANGO_REDIS_IGNORE_EXCEPTIONS`, every cache operation returns `None` — 100% cache miss, causing ~136M unnecessary DB queries on scalelab and significant p90/p99 latency regression.

- **How does this change address the issue?**
  `DABCacheWithFallback` delegates to the TCP-accessible `primary` Redis that the operator already configures, restoring working cache behavior on all deployment topologies.

## Type of Change

- [x] Configuration change

## Self-Review Checklist

### Prerequisites
An operator-deployed AAP environment with Redis running as a separate pod (not sidecar).

### Steps to Test
1. Deploy the gateway with this change
2. Exec into the gateway pod and run:
   ```
   python -c "import os,django; os.environ.setdefault('DJANGO_SETTINGS_MODULE','aap_gateway_api.settings'); django.setup(); from django.core.cache import caches; c=caches['default']; c.set('k','v'); print(c.get('k'))"
   ```
3. Verify the output is `v` (not `None`)

### Expected Results
Cache operations succeed via the TCP-accessible primary Redis. No `ConnectionError` on missing Unix socket.

## Summary by CodeRabbit

- **New Features**
  - Added fallback-enabled caching as the default cache backend.
  - Added a dedicated local Redis cache configuration.

- **Bug Fixes**
  - Improved cache resilience by enabling fallback behavior when the primary cache is unavailable.
  - Updated cache configuration validation to reflect the new defaults.


Dostonbek1 (approved):
## Code Review Summary

Restoring `DABCacheWithFallback` as `CACHES["default"]` is the right defensive fix for AAP-89439. Operator topologies never override `default`, so the Unix-socket `DABRedisCache` from af8d990c produced silent 100% cache misses. `DABCacheWithFallback` delegates to the TCP `primary` Redis the operator already configures, which is what 26-next does today.

**Files Reviewed**: 2 files
**Comments Posted**: 0 review comments

### Issues Found
- 0 security/safety issues
- 0 correctness/logic issues
- 0 code quality suggestions

### Overall Assessment
LGTM

### General Feedback
- `primary` and `fallback` aliases are unchanged, which is what `DABCacheWithFallback` and `/status` Redis health checks actually use.
- Nothing in-tree reads `CACHES["legacy"]`; renaming that slot to `redis_local` for sidecar opt-in is safe.
- Tests were updated to match the new aliases. `ANSIBLE_BASE_REDIS_AUTO_INVALIDATE` staying True is fine — it only activates when `DABRedisCache` is the active backend.

jeffh-oss (approved):

